PR #3, opened 2017-03-20
Introduce configurable MTU for Docker-encapsulated networking

oneswig:


oneswig:
Workflow -1 - need to work on better checks in the logic

ansible/roles/docker/templates/daemon.json.j2
```diff
@@ -1,8 +1,9 @@
 {
   "storage-driver": "devicemapper",
-   "storage-opts": [
-     "dm.thinpooldev=/dev/mapper/{{ docker_storage_volume_group | replace('-', '--') }}-{{ docker_storage_volume_thinpool | replace('-', '--') }}",
-     "dm.use_deferred_removal=true",
-     "dm.use_deferred_deletion=true"
-   ]
+  "mtu": 1450,
```
markgoddard: This should be configurable. The reduced MTU is caused by the iLab gateway. I've also been using MTU 1400 and found that anything greater wouldn't work for my test case of `curl https://galaxy.ansible.com`.

To try to remain DRY we could try to determine the MTU from the kayobe network config. You would do this as follows:

`{{ <network name> | net_mtu }}`

In our case, the relevant network is 'ilab':

`{{ 'ilab' | net_mtu }}`

or alternatively the more abstract network 'role' that is not alaska specific. I'm not sure whether this would always be correct though.

`{{ external_net_name | net_mtu }}`

ansible/roles/docker/tasks/main.yml
```diff
@@ -24,3 +24,16 @@
...
+    src: /etc/docker/daemon.json
+  register: docker_daemon_conf
+  become: True
+
+- name: Parse Docker daemon config JSON
+  set_fact:
+    docker_daemon_json: "{{ docker_daemon_conf['content'] | b64decode | from_json }}"
+
+- include: config.yml
+  when: "{{ docker_daemon_json['mtu'] }} != 1450 or {{ docker_daemon_json['storage-driver'] }} != 'devicemapper'"
```
markgoddard: I don't think we need any slurping here. We can just use the template module to determine if there were any changes. If there were not, the handler won't get triggered.

markgoddard (changes requested):
Looks good, needs a few changes.

ansible/roles/docker/tasks/main.yml
```diff
@@ -6,6 +6,12 @@
     append: yes
   become: True
 
+- name: Check Docker state with systemd
```
markgoddard: Name no longer accurate.

ansible/roles/docker/templates/daemon.json.j2
```diff
@@ -1,6 +1,6 @@
 {
   "storage-driver": "devicemapper",
-  "mtu": 1450,
+  "mtu": {{ external_net_name | net_mtu }},
```
markgoddard: Sorry, should have explained this in the previous comment. In general I've tried to make the role variables separate from the global variables to make the roles more self-contained and reusable. Sometimes they exist in the global variable space with the same name, sometimes they don't. It usually depends on whether they are shared between multiple plays and roles.

Typically I define all variables required by a role in the defaults files, prefixed by the name of the role to work around ansible's lack of namespacing. Often they're just unset or an empty list or dict. Then in the playbook where the role is used, if there isn't a 1-1 mapping from global to role variable we would do the mapping at that point. For this example, I would do:

ansible/roles/docker/defaults/main.yml:
`docker_daemon_mtu:`

ansible/roles/docker/templates/daemon.json.j2:
`"mtu": {{ docker_daemon_mtu }},`

ansible/docker.yml:
```
  roles:
     - role: docker
       docker_daemon_mtu: "{{ external_net_name | net_mtu }}"
```

markgoddard (approved):

PR #10, opened 2017-09-04
Setup Travis CI

markgoddard:
No more broken builds!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,28 @@
+---
+language: python
+python: "2.7"
+
+# Use the new container infrastructure
+sudo: false
+
+# Install ansible
+addons:
+  apt:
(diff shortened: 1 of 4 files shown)


markgoddard:
Note 'All checks have passed' above.

oneswig:
Travis might say it's passed but it doesn't look like a bed of roses in the log: https://travis-ci.org/stackhpc/kayobe/builds/271624127?utm_source=github_status&utm_medium=notification

markgoddard:
You mean the error messages in the unit test output? That's normal test output, although I agree it's not ideal.

oneswig (approved):
I'll take your word that it's testing (and interpreting the results) correctly.  It's a worthwhile direction of travel.

PR #12, opened 2017-09-04
Add static configuration of eth1 in development environment

markgoddard:
This is required in order to allow kayobe to reach the controller prior to
running the network configuration plays.


markgoddard:
You were right John - it hadn't been tested thoroughly enough and fell over a little later!

PR #13, opened 2017-09-07
Add support for Ubuntu control hosts

markgoddard:


oneswig (commented):
What's the motivation for Ubuntu support for the control host - for Ubuntu laptops acting as control host?  Looking forward to your Mac OS X patches...

The latter patches don't appear connected to Ubuntu portability issues?

markgoddard:
Yes, that's the idea - John and I can then use our laptops as the control host. I would love to add support for Mac OSX, but alas I'd have no means of testing it. Feel free to DIY...

markgoddard:
The latter patches aren't entirely relevant to Ubuntu support - correct, but they are still useful!

doc/source/installation.rst
```diff
@@ -5,26 +5,50 @@ Installation
...
+.. code-block:: console
+   :caption: CentOS
+
+   $ yum install -y git
+
+.. code-block:: console
+   :caption: Ubuntu
 
-    $ yum install -y git
+   $ apt install -y git
```
JohnGarbutt: I would be tempted to have a section that says, on Ubuntu you do these commands: ... But this works too.

markgoddard: I've reworked it a little, and simplified the package installation into a single command for each platform. WDYT?

PR #23, opened 2017-09-20
In latest Docker 17.05 restart-retires and unless-stopped are mutually exclusive. Remove restart-retries where used.

oneswig:


markgoddard (approved):
Lovely stuff.

markgoddard:
Is there a reason why we don't also need this for master (pike)?

PR #24, opened 2017-10-04
Wrong var name for redhat

ktibi:
Wrong name for variable.


oneswig (approved):
This makes sense.  Well spotted and thank you for sharing back.

PR #27, opened 2017-10-11
Junos support for ansible 2.3

markgoddard:
Configuration of juniper switches was broken following the switch to ansible 2.3. These changes fix that.

### DIFF
--- a/ansible/roles/junos-switch/tasks/main.yml
+++ b/ansible/roles/junos-switch/tasks/main.yml
@@ -3,6 +3,8 @@
 # not seem to be possible to use ansible_python_interpreter in combination with
 # delegate_to. This should be investigated in future as modification of system
 # packages via pip is not ideal.
+
+# For ansible 2.2, Juniper specific python packages are required.
 - name: Ensure python dependencies are installed
   pip:
     name: "{{ item }}"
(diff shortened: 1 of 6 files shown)


ansible/roles/junos-switch/tasks/main.yml
```diff
@@ -3,15 +3,29 @@
...
     - junos-eznc
     - jxmlease
+  when: ansible_version | version_compare('2.3', 'lt')
   become: True
 
+# For ansible 2.3+, only the NETCONF client ncclient is required. This must be
+# installed on the ansible control host where the ansible-connection subprocess
+# is executed.
+- name: Ensure python dependencies are installed
+  local_action:
```
oneswig, line 22: Just wondering about the semantic difference between local_action and delegate_to, and whether that has relevance to the comment beginning at line 2

oneswig (approved):
Looks good to me

ansible/roles/junos-switch/tasks/main.yml
markgoddard: Looks like delegate_to other than localhost is no longer supported in 2.3, so I'll change to local_action to be more explicit. I think that delegate_to: localhost *should* be equivalent to local_action though.

PR #29, opened 2017-10-13
Change name of element for diskimage-builder

ktibi:
Error when try to initiate image build ==>
/bifrost-base-source/bifrost-base-archive-stackhpc-pike/playbooks/roles/bifrost-create-dib-image/tasks/main.yml:121
 diskimage_builder.element_dependencies.MissingElementException: Element 'serial-console' not found
 diskimage-builder version 2.9.1

Check on https://docs.openstack.org/diskimage-builder/latest/elements/enable-serial-console/README.html

serial-console => enable-serial-console


markgoddard (approved):
Good catch. I have to admit we've not been through a seed deployment on pike yet - we've been reusing existing seeds to test the overcloud deployment.

PR #32, opened 2017-10-17
Physical network interface limit and --display

markgoddard:
* Allow limiting physical network interface configuration to subsets of interfaces.
* Add a --display argument to kayobe physical network configure


markgoddard:
@darrylweaver: switch interface configuration limiting.

darrylweaver (approved):
Looks good to me

PR #34, opened 2017-10-17
Added Prometheus node exporter role

jpds:


ansible/roles/node-exporter/README.md
```diff
@@ -0,0 +1,45 @@
...
+
+Example Playbook
+----------------
+
+The following playbook configures Node Exporter.
+
+    ---
+    - hosts: nodeexporter
+      roles:
+        - role: nodeexporter
```
markgoddard: node-exporter

ansible/roles/node-exporter/defaults/main.yml
```diff
@@ -0,0 +1,31 @@
...
+# Whether Node Exporter is enabled.
+nodeexporter_enabled: true
+
+# Service deployment definition.
+nodeexporter_services:
+  nodeexporter:
+    container_name: nodeexporter
+    enabled: "{{ nodeexporter_enabled }}"
+    image: "{{ nodeexporter_image_full }}"
+    command: /bin/node_exporter --collector.procfs=/host/proc --collector.sysfs=/host/sys --collector.filesystem.ignored-mount-points "^/(sys|proc|dev|host|etc)($|/)"
```
markgoddard, line 13: I don't see this being used.

ansible/roles/node-exporter/tasks/reconfigure.yml
```diff
@@ -0,0 +1 @@
+deploy.yml
```
markgoddard, line 1: Should this be a symlink? Alternatively, use `- include: deploy.yml`.

markgoddard (commented):
Mostly looks fine.

ansible/roles/node-exporter/tasks/reconfigure.yml
jpds: This one I copied from the opensm role.

markgoddard: Right, but it's a symlink in that role, and copying doesn't appear to have done the right thing (unless I'm missing something?).

markgoddard (approved):
Looks fine.

ansible/roles/node-exporter/defaults/main.yml
```diff
@@ -0,0 +1,31 @@
+---
+# Roughly follows kolla-ansible's service deployment patterns.
+
+# Whether Node Exporter is enabled.
+nodeexporter_enabled: true
```
markgoddard: I realise I've already approved this, but I looked again before merging and noticed this default. I'm not sure we want to enable this by default - it seems specific to the Verne deployment. Can we set it to false, and add an override in kayobe-config?

jpds: Sounds good to me, done.

markgoddard (approved):
Thanks.

PR #35, opened 2017-10-17
Add support for configuration of custom fluentd filters

markgoddard:
These may be placed in $KAYOBE_CONFIG_PATH/kolla/config/fluentd/filter.


ansible/roles/kolla-openstack/tasks/config.yml
```diff
@@ -41,6 +42,15 @@
     - "{{ kolla_extra_glance_path }}/*"
   when: kolla_extra_glance_path != None
 
+- name: Ensure extra fluentd filter configuration files exist
+  template:
+    src: "{{ item }}"
+    dest: "{{ kolla_node_custom_config_path }}/fluentd/filter/{{ item | basename }}"
```
dougszumski: If someone added a filter in kayobe, ran kayobe, removed it, and then re-ran kayobe, would the original filter persist here? In that case should we clear the folder before syncing the filters?

oneswig: Do we know that this is the only source of filters though - how would we know the filters we've removed from the filters others have added?  Do any filters come packaged in the Kolla fluentd container for example?

oneswig (commented):
very timely...

ansible/roles/kolla-openstack/tasks/config.yml
dougszumski: I was thinking about kayobe-config - hopefully everyone should check their filters into there in a timely fashion? That should address the first point. Kolla fluentd does provide filters, but those shouldn't be in the kayobe-config?

markgoddard: It's a valid point. In this case this should be the only source of filters, as the destination is the kolla-ansible config on the ansible control host (localhost), not the container config.

We can't really remove all the files, as that would break idempotence. I guess we could check for files that are in the destination and not the source.

dougszumski (approved):
Looks good to me. I haven't tested it yet.

dougszumski:
I just tested this and it works for me.

PR #37, opened 2017-10-18
 add version for ironic api

ktibi:
Fix a failure in the bifrost inventory script

### DIFF
--- a/ansible/overcloud-provision.yml
+++ b/ansible/overcloud-provision.yml
@@ -97,7 +97,6 @@
         docker exec bifrost_deploy
         bash -c '. env-vars &&
         export BIFROST_INVENTORY_SOURCE=ironic &&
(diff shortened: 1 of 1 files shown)


ansible/overcloud-provision.yml
```diff
@@ -96,6 +97,7 @@
         docker exec bifrost_deploy
         bash -c '. env-vars &&
         export BIFROST_INVENTORY_SOURCE=ironic &&
+        export OS_BAREMETAL_API_VERSION=1.34 &&
```
markgoddard: This one shouldn't be required - it's using the ironic client, not the openstack client.

markgoddard (changes requested):

ansible/overcloud-provision.yml
ktibi: done

markgoddard (approved):

PR #38, opened 2017-10-19
Fix when warnings & add no_log

markgoddard:
* Fix ansible warnings for when clauses with curly braces
* Add no_log to kolla_passwords module argument


dougszumski (commented):
Looks good to me. I didn't test but I ran it through ansible-lint and couldn't see any related complaints.

PR #41, opened 2017-10-23
Change kolla build configuration path variable

markgoddard:
Previously shared kolla-ansible's kolla_config_path, now uses
kolla_build_config_path.

Also adds a variable, config_path, which sets a base path for
configuration on the remote hosts.


dougszumski (approved):
LGTM.

PR #42, opened 2017-10-23
Use kolla-ansible's default neutron ML2 mechanism drivers

markgoddard:
The dependencies are not installed by default. The default list now
contains openvswitch only.


dougszumski (approved):

markgoddard:
This one requires changes to kayobe-config before merging, to ensure that genericswitch is explicitly enabled where necessary.

markgoddard:
Based on feedback from ktibi, the kolla-ansible defaults were found to be sensible for his use case, adding the l2population driver when using virtualisation.

PR #46, opened 2017-10-26
Add support for configuration of IP route tables and rules

markgoddard:


markgoddard:
Depends on https://github.com/michaelrigart/ansible-role-interfaces/pull/22.

PR #48, opened 2017-10-26
Add ceph config file for kolla

ktibi:


ansible/roles/kolla-openstack/vars/main.yml
```diff
@@ -77,3 +77,8 @@ kolla_openstack_custom_config:
     dest: "{{ kolla_node_custom_config_path }}/swift"
     patterns: "*"
     enabled: "{{ kolla_enable_swift }}"
+  # Ceph.
+  - src: "{{ kolla_extra_config_path }}/ceph"
+    dest: "{{ kolla_node_custom_config_path }}/ceph"
+    patterns: "*"
+    enabled: "{{ kolla_enable_ceph }}"
```
markgoddard: Minor nit: preserve alphabetical order.

markgoddard: I don't think we have kolla_enable_ceph yet, unless you have added it?

markgoddard:
Looks good, but depends on adding kolla_enable_ceph.

ktibi:
kolla_enable_ceph is already defined in etc/kayobe/kolla.yml no ?

markgoddard:
Defaults are set here: https://github.com/stackhpc/kayobe/blob/master/ansible/group_vars/all/kolla#L239

markgoddard (approved):
Great, thanks.

PR #50, opened 2017-10-27
Restrict prometheus node exporter to overcloud hosts running docker

markgoddard:
Also add some default global configuration to enable the service.

### DIFF
--- a/etc/kayobe/node-exporter.yml
+++ b/etc/kayobe/node-exporter.yml
@@ -1,0 +1,10 @@
+---
+###############################################################################
+# Prometheus node exporter configuration.
+
+# Whether Prometheus node exporter is enabled.
+#nodeexporter_enabled:
+
+###############################################################################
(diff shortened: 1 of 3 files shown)


etc/kayobe/node-exporter.yml
```diff
@@ -0,0 +1,10 @@
+---
+###############################################################################
+# Prometheus node exporter configuration.
+
+# Whether Prometheus node exporter is enabled.
+#nodeexporter_enabled:
```
jpds, line 6: Should this have a value?

markgoddard: It's just illustrate the variables that can be set - ansible/group_vars/node-exporter contains the defaults. This file forms the base of kayobe-config/etc/kayobe/node-exporter.yml.

jpds:
LGTM

PR #51, opened 2017-10-27
Set the Vagrant development VM's hostname

markgoddard:
By default the hostname will be localhost, which causes issues with RabbitMQ.


markgoddard:
FYI @weknowsoftware

PR #67, opened 2017-11-10
add CA conf for private registry

ktibi:
In my kayobe-conf I added :

- kayobe-config/etc/kayobe/docker/ca.crt
- kayobe-config/etc/kayobe/docker.yml :
```
# URL of docker registry
docker_registry: "{{ kolla_docker_registry }}"

# CA of docker registry
docker_registry_ca: "{{ kayobe_config_path }}/docker/certs/ca.crt"
```

### DIFF
--- a/ansible/group_vars/all/docker
+++ b/ansible/group_vars/all/docker
@@ -17,3 +17,9 @@ docker_storage_volume_thinpool_meta: docker-thinpoolmeta
 # Size of the docker storage metadata LVM volume (see lvol module size
 # argument).
 docker_storage_volume_thinpool_meta_size: 1%VG
+
+# URL of docker registry
+docker_registry:
+
+# CA of docker registry
+docker_registry_ca:
--- a/ansible/roles/docker/handlers/main.yml
+++ b/ansible/roles/docker/handlers/main.yml
@@ -4,3 +4,9 @@
     name: docker
     state: restarted
   become: True
+
+- name: reload docker service
+  service:
+    name: docker
+    state: reloaded
+  become: True
--- a/ansible/roles/docker/tasks/config.yml
+++ b/ansible/roles/docker/tasks/config.yml
@@ -5,3 +5,11 @@
     dest: /etc/docker/daemon.json
   become: True
   notify: restart docker service
+
+- name: Ensure the CA file for private registry exists
+  copy:
+    src: "{{ docker_registry_ca }}"
+    dest: "/etc/docker/certs.d/{{ docker_registry }}/ca.crt"
+  become: True
+  when: docker_registry is not none and docker_registry_ca is not none
+  notify: reload docker service
--- a/ansible/roles/docker/defaults/main.yml
+++ b/ansible/roles/docker/defaults/main.yml
@@ -25,5 +25,11 @@ docker_storage_thinpool_autoextend_threshold: 80
 # Percentage by which to extend thin-provisioned docker storage volumes.
 docker_storage_thinpool_autoextend_percent: 20
 
+# URL of docker registry
+docker_registry:
+
+# CA of docker registry
+docker_registry_ca:
+
 # MTU to pass through to containers not using net=host
 docker_daemon_mtu: 1500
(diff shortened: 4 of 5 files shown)


ansible/roles/docker/tasks/config.yml
```diff
@@ -5,3 +5,11 @@
     dest: /etc/docker/daemon.json
   become: True
   notify: restart docker service
+
+- name: Ensure the CA file for private registry exists
+  copy:
+    src: "{{ docker_registry_ca }}"
+    dest: "/etc/docker/certs.d/{{ docker_registry }}/ca.crt"
+  become: True
+  when: kolla_docker_registry is not none and docker_registry_ca is not none
```
markgoddard: s/kolla_//

ansible/roles/docker/tasks/config.yml
```diff
@@ -5,3 +5,11 @@
     dest: /etc/docker/daemon.json
   become: True
   notify: restart docker service
+
+- name: Ensure the CA file for private registry exists
+  copy:
+    src: "{{ docker_registry_ca }}"
+    dest: "/etc/docker/certs.d/{{ docker_registry }}/ca.crt"
```
markgoddard, line 12: Would it be possible to need more than one CA? Would this be more flexible:

```
# Dict mapping hostnames of registries to an x.509 certificate authority file for that registry. 
docker_registry_ca: {}
```
Example:
```
docker_registry_ca:
  my-registry: "<CA file>"
```

ansible/roles/docker/handlers/main.yml
```diff
@@ -4,3 +4,9 @@
     name: docker
     state: restarted
   become: True
+
+- name: reload docker service
```
markgoddard, line 8: Should we always use reload?

markgoddard (changes requested):

ansible/roles/docker/tasks/config.yml
ktibi: normally it's one CA per registry. so one directory by registry.

ansible/roles/docker/handlers/main.yml
ktibi: yes reload load new CA

ansible/roles/docker/tasks/config.yml
markgoddard: Right, but there could be multiple registries in use by a single docker engine?

ansible/roles/docker/handlers/main.yml
markgoddard: I mean in the previous task, we restart docker. Should we always do a reload instead of restart? Or is reload not always enough?

ansible/roles/docker/tasks/config.yml
ktibi: hum yes you can, but for now, kayobe and kolla use one registry. docker_registry is not a dict mapping.

ansible/roles/docker/handlers/main.yml
ktibi: If you change conf of docker I don't know if reload it's enough

ansible/roles/docker/tasks/config.yml
markgoddard: That's true, although I think it's still valid as the kayobe services may change but we don't want to keep changing docker. I guess for public registries it won't be necessary, and it's less likely that we would need multiple private registries. We can go this route until we need to change.

ansible/group_vars/all/docker
```diff
@@ -17,3 +17,9 @@ docker_storage_volume_thinpool_meta: docker-thinpoolmeta
 # Size of the docker storage metadata LVM volume (see lvol module size
 # argument).
 docker_storage_volume_thinpool_meta_size: 1%VG
+
+# URL of docker registry
+docker_registry:
+
+# CA of docker registry
+docker_registry_ca:
```
markgoddard, line 25: These two should be added to etc/kayobe/docker.yml.

markgoddard (changes requested):

ansible/group_vars/all/docker
ktibi: done

markgoddard (approved):

PR #69, opened 2017-11-13
Add creation for directory CA docker

ktibi:


ansible/roles/docker/tasks/config.yml
```diff
@@ -6,6 +6,13 @@
   become: True
   notify: restart docker service
 
+- name: Ensure the path for CA file for private registry exists
+  file:
+    path: "/etc/docker/certs.d/{{ docker_registry }}"
```
markgoddard, line 11: Does it need to have any particular mode and/or ownership?

ansible/roles/docker/tasks/config.yml
```diff
@@ -6,6 +6,13 @@
...
+    state: directory
+  become: True
+  when: kolla_docker_registry is not none and docker_registry_ca is not none
```
markgoddard: Should be docker_registry.

markgoddard (changes requested):

ansible/roles/docker/tasks/config.yml
ktibi: no works with become: true

markgoddard (approved):

Issue #71, opened 2017-11-13
kolla is attempting to read in in an ini file as YAML during `kayobe overcloud host configure`


autonomouse:
kolla is attempting to read in in an ini file as YAML during `kayobe overcloud host configure`

```
        Bootstraping servers : ansible-playbook -i /vagrant/kayobe-config_staging/etc/kolla/inventory/overcloud -e @/vagrant/kayobe-config_staging/etc/kolla/globals.yml -e @/vagrant/kayobe-config_staging/etc/kolla/passwords.yml -e CONFIG_DIR=/vagrant/kayobe-config_staging/etc/kolla  --vault-password-file=/root/kayobe-venv/bin/kayobe-vault-password-helper -e ansible_user=stack -e action=bootstrap-servers /vagrant/kayobe/venvs/kolla-ansible/share/kolla-ansible/ansible/kolla-host.yml 
        ERROR! Attempted to read "/vagrant/kayobe-config_staging/etc/kolla/inventory/overcloud" as YAML: Syntax Error while loading YAML.
        ​
        ​
        The error appears to have been in '/vagrant/kayobe-config_staging/etc/kolla/inventory/overcloud': line 15, column 1, but may
        be elsewhere in the file depending on the exact syntax problem.
        ​
        The offending line appears to be:
        ​
        # These hostnames must be resolvable from your deployment host
        stg-vircon0001 ansible_host=10.114.143.2
        ^ here
        ​
        Attempted to read "/vagrant/kayobe-config_staging/etc/kolla/inventory/overcloud" as ini file: /vagrant/kayobe-config_staging/etc/kolla/inventory/overcloud:259: Section [nova-compute-ironic:children] includes undefined group: stg-vircon0002 
        Command failed ansible-playbook -i /vagrant/kayobe-config_staging/etc/kolla/inventory/overcloud -e @/vagrant/kayobe-config_staging/etc/kolla/globals.yml -e @/vagrant/kayobe-config_staging/etc/kolla/passwords.yml -e CONFIG_DIR=/vagrant/kayobe-config_staging/etc/kolla  --vault-password-file=/root/kayobe-venv/bin/kayobe-vault-password-helper -e ansible_user=stack -e action=bootstrap-servers /vagrant/kayobe/venvs/kolla-ansible/share/kolla-ansible/ansible/kolla-host.yml 
        kolla-ansible bootstrap-servers exited 1

```

Yet if I run that command manually, it works!

[...]

PR #72, opened 2017-11-13
(potential) fix for https://github.com/stackhpc/kayobe/issues/71

autonomouse:


markgoddard (commented):
This is basically the right approach. We don't need kolla_nova_compute_ironic_group to be configurable, so that variable should be removed and hard coded to nova in the inventory.

ansible/roles/kolla-ansible/defaults/main.yml
```diff
@@ -210,6 +210,8 @@ kolla_openstack_logging_debug:
 # Nova options
 #######################
 
+kolla_nova_compute_ironic_host:
```
markgoddard: Needs a comment similar to the one below, and the commented out version in etc/kayobe/nova.yml.

ansible/roles/kolla-ansible/templates/overcloud-services.j2
```diff
@@ -37,11 +37,16 @@ nova
...
-[nova-compute-ironic:children]
 # NOTE: HA for nova-compute services with ironic is still an experimental
 # feature. Provide the option to use a single compute host, even when multiple
 # controllers are in use.
+{% if kolla_nova_compute_ironic_host is not none %}
+[nova-compute-ironic]
+{{ kolla_nova_compute_ironic_host }}
+{% else %}
+[nova-compute-ironic:children]
 {{ kolla_nova_compute_ironic_group }}
```
markgoddard: The default group doesn't need to be configurable - can we set this to just `nova`, then remove all references to kolla_nova_compute_ironic_group?

markgoddard (changes requested):

ansible/roles/kolla-ansible/templates/overcloud-services.j2
autonomouse: done

markgoddard (approved):
Looks good, thanks

PR #74, opened 2017-11-16
add ldap conf for keystone

ktibi:


markgoddard:
Thanks for this @ktibi. I had a thought about this - can we set `recurse: True`  in the two `find` tasks to achieve this and a generally more flexible configuration mechanism?

https://github.com/stackhpc/kayobe/blob/master/ansible/roles/kolla-openstack/tasks/config.yml#L58

ansible/group_vars/all/kolla
```diff
@@ -271,6 +271,7 @@ kolla_enable_heat: "yes"
 kolla_enable_horizon: "yes"
 kolla_enable_ironic: "yes"
 kolla_enable_kibana: "{{ 'yes' if kolla_enable_central_logging | bool else 'no' }}"
+kolla_enable_keystone_ldap: "no"
```
markgoddard: I don't think we need this - if the domain files are present we can copy them into place, if not we do nothing.

markgoddard (changes requested):

ansible/group_vars/all/kolla
```diff
@@ -281,6 +281,10 @@ kolla_enable_skydive: "no"
 kolla_enable_swift: "no"
 
 ###############################################################################
+## Kolla horizon configuration.
+kolla_horizon_keystone_multidomain: "False"
```
markgoddard: We are just starting to use this flag as well. Can we just keep it outside of kayobe though? If we have to own all configuration options it will be hard to maintain.

You can add this by creating `$KAYOBE_CONFIG_PATH/kolla/globals.yml` and adding it there.

markgoddard (changes requested):

ktibi:
ok with you about new kolla options

markgoddard (approved):
Great, thanks

Issue #16, opened 2017-09-20
Improve container image build workflow


markgoddard:
The existing container image build workflow has a number of areas where it could be improved.

* Cleaner separation of the build process from the running system. Seed images are currently built on the seed, and control plane images on all of the controllers. Ideally, these would be decoupled, allowing for images to be built and tested elsewhere. The addition of e.g. [seed-container-builder] and [overcloud-container-builder] groups could help here. It should still be possible to build locally on the seed or controller hosts, as this is convenient for development and testing.

* Documentation of the workflow, use of registries, and a build, push, pull model.

markgoddard:
@JohnGarbutt

JohnGarbutt:
For extra context, my thinking on this is quite influenced by:
https://12factor.net/

I think it would be good to have clear "release" artifacts out of a build stage, all very separate from the configuration for a given environment. So in the CI system (once we have it) we test the exact same version of all decencies we have in production, the same docker images, same kayobe venv, ideally the same controller images, etc.

In theory we should be able to run the deploy without any internet connectivity, not because that is important, but because that means we know we are running the exact same code in dev, test and production.

Separating out a builders group seems a good step in the right direction.

I would be tempted to lean much less on the Docker registry, and more on a tarball based build artifact style distribution. The tarball can then be used to populate the deploy local Docker registry for pushing out to all the controllers. I like the idea of first doing a checksum of the tarball before importing into the local docker registry, using that checksum as a tag on all the docker images, etc.

markgoddard:
Thanks John, those are all good points.

On the Docker registry, do you mean we should be leaning less on *the* docker registry (dockerhub) or *any* docker registry (which could be local)? With a local registry we could the tag built images with a version/build number.

In the tarball method, it seems to add some complexity over the registry. Where would these tarballs be stored? How do we store metadata? We also lose the layering of docker images, which may or may not be useful.

JohnGarbutt:
Keeping the registry probably is simpler. I think they key bit I like is ensuring there are known good builds you can easily pull. Ideally using the same build for almost all users.

markgoddard:
Migrated to storyboard: https://storyboard.openstack.org/#!/story/2001669

PR #81, opened 2017-11-21
Improve container image build workflow

markgoddard:
Container images are now built by hosts in the `[container-image-builders]` group, which defaults to the seed.

Related: #16


ktibi (approved):
goods. Need to update doc with : 

Seed build all images by default.

PR #84, opened 2017-11-27
add storage group and ceph tag

ktibi:


ansible/ceph.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Ensure tag CEPH configuration is applied
+  hosts: overcloud
```
markgoddard: This will match all hosts in the overcloud group, which will include monitoring hosts, and any custom groups in overcloud as well. These hosts won't currently have a ceph_disks variable defined and will fail.

We can either define a default value in group_vars/all for ceph_disks, or use `ceph_disks | default([])` here.

ansible/ceph.yml
```diff
@@ -0,0 +1,28 @@
+---
+- name: Ensure tag CEPH configuration is applied
```
markgoddard: How about:

Ensure block devices for use by Ceph are labelled

ansible/ceph.yml
```diff
@@ -0,0 +1,28 @@
...
+  hosts: overcloud
+  tasks:
+    - name: Check the presence of a partition on the disks
+      become: True
+      parted:
+        device: "{{ item }}"
+      with_items: "{{ ceph_disks }}"
+      register: "disk_info"
```
markgoddard: nit: Empty line between tasks.

ansible/ceph.yml
```diff
@@ -0,0 +1,28 @@
...
+    - name: Fail if the CEPH physical disks have already a partition
```
markgoddard: Does this make the playbook not idempotent? Ideally we should be able to run this playbook multiple times. Can we not just check that the partition is as expected?

ansible/ceph.yml
markgoddard: Also, minor nit: Ceph isn't capitalised.

markgoddard (changes requested):
Thanks @ktibi.

ansible/ceph.yml
ktibi: I'll try to modify the check 
if partition name is already kolla tag ==> no fail

markgoddard:
Should this be integrated with `kayobe overcloud host configure`?

markgoddard:
Can we also make the playbook name a bit more specific - e.g. `ceph-block-devices.yml`?

ansible/ceph.yml
```diff
@@ -0,0 +1,28 @@
...
+      when: item.partitions | length > 0
+      loop_control:
+        label: "{{item.item}}"
+    - name: Create tag partition for CEPH OSD
+      become: True
+      parted:
+        device: "{{ item }}"
+        number: 1
+        label: gpt
+        name: "KOLLA_CEPH_OSD_BOOTSTRAP"
```
markgoddard: Thinking ahead, how would this be extended to support an external journal drive or a cache tier? https://docs.openstack.org/kolla-ansible/ocata/ceph-guide.html

ktibi: yes, I can create second list with a test 
if journal list is empty, use standard conf

ktibi:
other points : 
- for now, parted module need ansible 2.4 to be able to use feature about partition name. Maybe we can use library, or shell command. But with a view to upgrade one day ansible 2.3 to 2.4, use new parted feature is better.
- Need to add partition deletion when deprovision because ceph can be deploy if disks have always the old partition.

ansible/ceph-block-devices.yml
```diff
@@ -0,0 +1,30 @@
+---
+- name: Ensure block devices for use by Ceph are labelled
+  hosts: overcloud
+  tasks:
+    - name: Check the presence of a partition on the disks
+      become: True
+      parted:
+        device: "{{ item }}"
+      with_items: "{{ ceph_disks }} | default([])"
```
markgoddard: |default([]) needs to be inside }}.

ansible/ceph-block-devices.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: Create tag partition for Ceph OSD
+      become: True
+      parted:
+        device: "{{ item }}"
+        number: 1
+        label: gpt
+        name: "KOLLA_CEPH_OSD_BOOTSTRAP"
+        state: present
+      with_items: "{{ ceph_disks }} | default([])"
```
markgoddard: As above

ansible/ceph-block-devices.yml
```diff
@@ -0,0 +1,30 @@
...
+
+    - name: Fail if the Ceph physical disks have already a partition
+      fail:
+        msg: >
+          The physical disk {{ item.item }} have already a partition.
+          Ensure that each disk in 'ceph_disks' have not a partition.
+      with_items: "{{ disk_info.results }}"
+      when: item.partitions | length > 0 and item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP'
+      loop_control:
+        label: "{{item.item}}"
```
markgoddard: Interesting, I haven't seen this used before.

ansible/ceph-block-devices.yml
```diff
@@ -0,0 +1,30 @@
...
+      parted:
+        device: "{{ item }}"
+      with_items: "{{ ceph_disks }} | default([])"
+      register: "disk_info"
+
+    - name: Fail if the Ceph physical disks have already a partition
+      fail:
+        msg: >
+          The physical disk {{ item.item }} have already a partition.
+          Ensure that each disk in 'ceph_disks' have not a partition.
```
markgoddard: Grammar nit:

The physical disk {{ item.item }} already has a partition.
Ensure that each disk in 'ceph_disks' does not have any partitions.

ansible/ceph-block-devices.yml
markgoddard: The workaround I suggested previously was essentially this:

```
with_items: "{{ disk_info.results }}"
when: disk_info.partitions | length == 0
```

markgoddard (changes requested):

ansible/ceph-block-devices.yml
ktibi: the job is not idempotent with that. If you have one partition but the good partition, job fail :/

ansible/ceph-block-devices.yml
```diff
@@ -0,0 +1,30 @@
...
+      with_items: "{{ disk_info.results }}"
+      when: item.partitions | length > 0 and item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP'
+      loop_control:
+        label: "{{item.item}}"
+
+    - name: Create tag partition for Ceph OSD
+      become: True
+      parted:
+        device: "{{ item }}"
+        number: 1
```
oneswig: We have a use case with NVMe drives for Ceph where there are 4 OSDs per NVMe.
It is an assumption that Ceph needs one OSD partition per drive because this makes sense for HDDs.
I am not sure if Kolla makes the same assumption.

ktibi: This tasks create a partition for tag the disk. 
After that, kolla remove this partition and create two partitions : one for data, one for journal.
Each OSD container manage one HDD.

oneswig: Ah OK, so the limitation is in kolla-ansible.
Do you want an external journal disk, like in https://docs.openstack.org/kolla-ansible/pike/ceph-guide.html#using-an-external-journal-drive

ktibi: I will add second list with journal disk, and if list is not empty, kayobe will tag disk with journal tag.

But I need to check with kolla ceph core team if upgrade to luminous is in roadmap for queens. Because with new storage method, journal can be on data disk without performance drop.

ansible/ceph-block-devices.yml
markgoddard: If there is one partition, the task will not run because `disk_info.partitions | length == 0` is not true.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
+---
+- name: Check the presence of a partition on the disks
+  become: True
+  parted:
+    device: "{{ item }}"
+  with_items: "{{ ceph_disks  | default([]) }} + {{ ceph_journal_disks  | default([]) }}"
```
markgoddard: Don't need the defaults here, as they are set in the defaults file.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+  register: "disk_info"
+
+- name: Fail if the Ceph data disks have already a partition
+  fail:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_info.results }}"
+  when: item.partitions | length > 0 and (item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP' or item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO_J' or item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO')
```
markgoddard: This line is quite long. You can format `and` operations using a list, and use a variable to simplify the `or`s:
```
when:
  - item.partitions | length > 0
  - item.partitions.0.name not in valid_labels
  vars:
    valid_labels:
      - KOLLA_CEPH_OSD_BOOTSTRAP
      - KOLLA_CEPH_OSD_BOOTSTRAP_FOO_J
      - KOLLA_CEPH_OSD_BOOTSTRAP_FOO
```

However, I think this is actually not quite right. The suffix does not have to be FOO, it can be anything - it is used by kolla-ansible to match the journals to their disks: https://docs.openstack.org/kolla-ansible/ocata/ceph-guide.html.  (code: https://github.com/openstack/kolla-ansible/blob/bf0e8a8d52a7c862f97c5b6baa0863571ec3313f/ansible/roles/ceph/tasks/bootstrap_osds.yml and https://github.com/openstack/kolla/blob/master/docker/kolla-toolbox/find_disks.py).

How about:
```
when:
  - item.partitions | length > 0
  - not item.partitions.0.name.startswith('KOLLA_CEPH_OSD_BOOTSTRAP')
```

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- name: Create tag partition for Ceph OSD
+  become: True
+  parted:
+    device: "{{ item }}"
+    number: 1
+    label: gpt
+    name: "{% if ceph_external_journal_enable %} 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO' {% else %} 'KOLLA_CEPH_OSD_BOOTSTRAP' {% endif %}"
+    state: present
+  with_items: "{{ ceph_disks | default([]) }}"
```
markgoddard: Default not required

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+
+- name: Create tag partition for Ceph external journal
+  become: True
+  parted:
+    device: "{{ item }}"
+    number: 1
+    label: gpt
+    name: "KOLLA_CEPH_OSD_BOOTSTRAP_FOO_J"
+    state: present
+  with_items: "{{ ceph_disks | default([]) }}"
```
markgoddard: Default not required. Also, should be ceph_journal_disks.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+  loop_control:
+    label: "{{item.item}}"
+
+- name: Create tag partition for Ceph OSD
+  become: True
+  parted:
+    device: "{{ item }}"
+    number: 1
+    label: gpt
+    name: "{% if ceph_external_journal_enable %} 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO' {% else %} 'KOLLA_CEPH_OSD_BOOTSTRAP' {% endif %}"
```
markgoddard: If we are using external journals, we need to generate unique suffixes for each disk/journal pair - not FOO. How do we know which journal goes with each disk? Should we just zip the lists together? If we have:

```
ceph_disks: [sda, sdb]
ceph_journal_disks: [sdc, sdd]
```

Then is sdc the journal for sda? Any scheme may need to consider what happens if we add a disk/journal to these lists.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+- name: Fail if the Ceph data disks have already a partition
+  fail:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_info.results }}"
+  when: item.partitions | length > 0 and (item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP' or item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO_J' or item.partitions.0.name != 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO')
+  loop_control:
+    label: "{{item.item}}"
+
```
markgoddard: Do we need to check that `ceph_disks | length == ceph_journal_disks | length`?

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+- name: Create tag partition for Ceph external journal
+  become: True
+  parted:
+    device: "{{ item }}"
+    number: 1
+    label: gpt
+    name: "KOLLA_CEPH_OSD_BOOTSTRAP_FOO_J"
+    state: present
+  with_items: "{{ ceph_disks | default([]) }}"
+  when: ceph_external_journal_enable
```
markgoddard: Do we need this flag? We could just set `ceph_journal_disks: []`

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,39 @@
...
+    device: "{{ item }}"
+    number: 1
+    label: gpt
+    name: "{% if ceph_external_journal_enable %} 'KOLLA_CEPH_OSD_BOOTSTRAP_FOO' {% else %} 'KOLLA_CEPH_OSD_BOOTSTRAP' {% endif %}"
+    state: present
+  with_items: "{{ ceph_disks | default([]) }}"
+
+- name: Create tag partition for Ceph external journal
+  become: True
+  parted:
```
markgoddard: Do we need to ensure that parted is installed?

ansible/group_vars/all/compute
```diff
@@ -101,6 +101,18 @@ compute_lvm_group_data_lv_docker_volumes_size: 75%VG
...
+# Compute node Ceph configuration.
+
+# List of Ceph disks.
+compute_ceph_disks: []
+
+# List of Ceph journal disks.
+compute_ceph_journal_disks: []
+
+# Whether to enable external journal.
+compute_ceph_journal_enable:
```
markgoddard: This should have a default value (although I suggest below that it is not required).

ansible/group_vars/all/storage
```diff
@@ -0,0 +1,127 @@
...
+# List of networks to which storage nodes are attached.
+storage_network_interfaces: >
+  {{ (storage_default_network_interfaces +
+      storage_extra_network_interfaces) | unique | list }}
+
+# List of default networks to which storage nodes are attached.
+storage_default_network_interfaces: >
+  {{ [provision_oc_net_name,
+      internal_net_name,
+      storage_net_name] | unique | list }}
```
markgoddard: Also storage_mgmt_net_name?

ansible/group_vars/all/storage
```diff
@@ -0,0 +1,127 @@
...
+  mntp: /var/lib/docker/volumes
+
+# Size of docker volumes LVM backing volume.
+storage_lvm_group_data_lv_docker_volumes_size: 75%VG
+
+# Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
+storage_lvm_group_data_lv_docker_volumes_fs: ext4
+
+###############################################################################
+# Storage node CEPH configuration.
```
markgoddard: nit: Ceph

etc/kayobe/compute.yml
```diff
@@ -83,6 +83,12 @@
 # Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
 #compute_lvm_group_data_lv_docker_volumes_fs:
 
+###############################################################################
+# Compute node CEPH configuration.
+
+# List of CEPH disks.
+# compute_ceph_disks:
```
markgoddard: Needs other variables.

etc/kayobe/controllers.yml
```diff
@@ -86,6 +86,12 @@
 # Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
 #controller_lvm_group_data_lv_docker_volumes_fs:
 
+###############################################################################
+# Controller node CEPH configuration.
+
+# List of CEPH disks.
+# controller_ceph_disks:
```
markgoddard: Needs other variables.

etc/kayobe/controllers.yml
```diff
@@ -86,6 +86,12 @@
...
+
 ###############################################################################
 # Controller node sysctl configuration.
```
markgoddard: Need to add etc/kayobe/storage.yml

markgoddard (changes requested):

ansible/group_vars/all/storage
ktibi: So compute too, because ceph-mon always on controllers.

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: Can I use : 
{% if ceph_journal_disks == [] %} for L42 ?

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: ceph_disks: [sda, sdb]
ceph_journal_disks: 
  - { device: sdc, osd : sda }
  - { device: sdd, osd : sdb }

ansible/roles/kolla-ceph/tasks/config.yml
markgoddard: This is fine:

`{% if ceph_journal_disks %}`

ansible/group_vars/all/storage
markgoddard: Sorry, I don't follow you here. I think we need to add storage_mgmt_net_name to this list, for the Ceph replication network.

ansible/group_vars/storage/bios
```diff
@@ -0,0 +1,7 @@
+---
+###############################################################################
+# Storage node BIOS configuration.
+
+# Dict of monitoring node BIOS options. Format is same as that used by
```
markgoddard: monitoring

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,56 @@
...
+  loop_control:
+    label: "{{item.item}}"
+
+- name: Check if the number of journals is correct
+  fail:
+    msg: >
+      Wrong number of journals. The number of journals need to be same
+      of number of OSD disks.
+  when: 
+    - ceph_disks | length != ceph_journal_disks | length
```
markgoddard: Is it valid to have some disks with journals and some without?

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,56 @@
...
+- name: Create tag partition for Ceph external journal
+  become: True
+  parted:
+    device: "{{ item.device }}"
+    number: 1
+    label: gpt
+    name: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.osd | basename }}_J"
+    state: present
+  with_items: "{{ ceph_journal_disks }}"
+  when: ceph_external_journal_enable
```
markgoddard: If this is run after the initial partition creation when Ceph is using the disks, will it work correctly and keep data intact?

etc/kayobe/compute.yml
```diff
@@ -83,6 +83,18 @@
 # Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
 #compute_lvm_group_data_lv_docker_volumes_fs:
 
+###############################################################################
+# Compute node Ceph configuration.
+
+# List of Ceph disks.
+#compute_ceph_disks: []
```
markgoddard: Nit: normally I don't include the default values here.

etc/kayobe/storage.yml
```diff
@@ -0,0 +1,113 @@
...
+#storage_lvm_group_data_lv_docker_volumes_size:
+
+# Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
+#storage_lvm_group_data_lv_docker_volumes_fs:
+
+###############################################################################
+# Storage node Ceph configuration.
+
+# List of Ceph disks.
+#storage_ceph_disks: []
```
markgoddard: Nit: normally I don't include the default values here.

ansible/roles/kolla-ceph/defaults/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+
+# List of Ceph disks.
+ceph_disks: []
+
+# List of Ceph external journal disks.
+ceph_journal_disks: []
+
+# Whether to enable external journal.
+ceph_external_journal_enable: False
```
markgoddard: I'm still not convinced this variable is required. I don't feel strongly enough about it to block the review though.

ansible/group_vars/all/compute
```diff
@@ -101,6 +101,18 @@ compute_lvm_group_data_lv_docker_volumes_size: 75%VG
 # Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
 compute_lvm_group_data_lv_docker_volumes_fs: ext4
 
+###############################################################################
+# Compute node Ceph configuration.
+
+# List of Ceph disks.
+compute_ceph_disks: []
+
+# List of Ceph journal disks.
```
markgoddard: This needs to describe the variable's data format as it is now non-trivial.

markgoddard (changes requested):

ansible/group_vars/all/storage
ktibi: Sorry my comment is not very clear. 
For now, I use a ceph cluster with osd on compute node. storage_mgmt_net_name is not on compute node in kayobe : https://github.com/stackhpc/kayobe/blob/master/ansible/group_vars/all/compute#L18

I think this network is just for ceph-mon. Because replication works on my cluster :/

markgoddard: Kayobe configures cluster_interface in kolla-ansible using this network. That interface is used for Ceph replication: https://docs.openstack.org/kolla-ansible/pike/user/production-architecture-guide.html#interface-configuration. Perhaps you have configured storage_mgmt_net_name == storage_net_name? I think that in general we do need storage_mgmt_net_name on the storage nodes.

ktibi: yes I used compute_network_interfaces for add both networks on my computes nodes.

ansible/group_vars/all/compute
ktibi: in doc or comment before var ?

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: If disk have already partition, the role fail. 

So I think I need to change with a register and add a condition on other tasks to skip itself if ceph already deployed.

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: each OSD need to have a journal ! so if this case, if you use external journal you need to have exactly the same number.

markgoddard: You can't mix local and external journals?

ansible/roles/kolla-ceph/tasks/config.yml
markgoddard: How do you know if Ceph is already deployed?

ansible/group_vars/all/compute
markgoddard: Both, ideally. At least in the comment.

ansible/group_vars/all/compute
```diff
@@ -101,6 +101,16 @@ compute_lvm_group_data_lv_docker_volumes_size: 75%VG
...
 compute_lvm_group_data_lv_docker_volumes_fs: ext4
 
+###############################################################################
+# Compute node Ceph configuration.
+
+# List of Ceph disks.
+# The format is a list of dict like :
+# - { osd: "/dev/sdb", journal: "/dev/sdc" }
+# - { osd: "/dev/sdd" }
+# Journal variable is not mandatory.
```
markgoddard, line 111: That's better (both format and the comment).

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
...
+    state: installed
+  become: True
+
+- name: Check the presence of a partition on the disks
+  become: True
+  parted:
+    device: "{{ item.osd }}"
+  with_items: "{{ ceph_disks }}"
+  register: "disk_info"
+  when: ceph_disks is defined
```
markgoddard: ceph_disks should always be defined as it's in defaults.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
+---
+- name: Ensure required packages are installed
+  yum:
+    name: parted
+    state: installed
+  become: True
```
markgoddard: `when: ceph_disks | length > 0`

ansible/roles/kolla-ceph/tasks/config.yml
markgoddard: Do we also need to check the journals?

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Fail if the Ceph data disks have already a partition
+  debug:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_info.results }}"
+  when: 
+    - item.partitions | length > 0 
+    - not item.partitions.0.name.startswith('KOLLA_CEPH_OSD_BOOTSTRAP')
+    - ceph_disks is defined
```
markgoddard: As above

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Check the presence of a partition on the disks
+  become: True
+  parted:
+    device: "{{ item.osd }}"
+  with_items: "{{ ceph_disks }}"
+  register: "disk_info"
+  when: ceph_disks is defined
+
+- name: Fail if the Ceph data disks have already a partition
+  debug:
```
markgoddard: Should this be fail? If not, need to rename the task and ceph_fail variable.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
...
+- name: Create tag partition for Ceph OSD
+  become: True
+  parted:
+    device: "{{ item.osd }}"
+    number: 1
+    label: gpt
+    name: "{% if item.journal is defined %} 'KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.osd | basename }}' {% else %} 'KOLLA_CEPH_OSD_BOOTSTRAP' {% endif %}"
+    state: present
+  with_items: "{{ ceph_disks }}"
+  when: not ceph_fail.skipped
```
markgoddard: Are you trying to use this to only create the partitions initially? I think there are some problems.

* It would be better to check each disk individually, then disks could be added. You can do this by iterating over ceph_fail.results, and checking item.skipped.
* Due to `item.partitions | length > 0`, this will not create a partition on a disk with no existing partitions.

Maybe it would be better to iterate over disk_info.results here.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
...
+
+- name: Fail if the Ceph data disks have already a partition
+  debug:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_info.results }}"
+  when: 
+    - item.partitions | length > 0 
+    - not item.partitions.0.name.startswith('KOLLA_CEPH_OSD_BOOTSTRAP')
```
markgoddard: We now know what the partition name should be, so we can match exactly:

`{% if item.journal is defined %} 'KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.osd | basename }}' {% else %} 'KOLLA_CEPH_OSD_BOOTSTRAP' {% endif %}`

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,52 @@
...
+  parted:
+    device: "{{ item.journal }}"
+    number: 1
+    label: gpt
+    name: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.osd | basename }}_J"
+    state: present
+  with_items: "{{ ceph_disks }}"
+  when: 
+    - item.journal is defined
+    - not ceph_fail.skipped
```
markgoddard: As above.

ansible/roles/kolla-ceph/tasks/config.yml
markgoddard: What possibilities are there?
* No partitions -> create a partition
* One or more partitions, first one has the correct label -> do nothing
* One or more partitions, first one has an incorrect label -> error?

Does it matter if there is more than one partition? What does the disk look like after kolla-ansible has run?

markgoddard (changes requested):

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: I change fail to debug because if anyone want to manualy tag disks. I'll rename var

markgoddard: Manual tagging should be the same as if kayobe had already tagged the disks and the host configure command was run again, right?

etc/kayobe/storage.yml
```diff
@@ -0,0 +1,107 @@
...
+# Size of docker volumes LVM backing volume.
+#storage_lvm_group_data_lv_docker_volumes_size:
+
+# Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
+#storage_lvm_group_data_lv_docker_volumes_fs:
+
+###############################################################################
+# Storage node Ceph configuration.
+
+# List of Ceph disks.
```
markgoddard, line 89: Need the improved comment here.

etc/kayobe/controllers.yml
```diff
@@ -86,6 +86,12 @@
 # Filesystem for docker volumes LVM backing volume. ext4 allows for shrinking.
 #controller_lvm_group_data_lv_docker_volumes_fs:
 
+###############################################################################
+# Controller node Ceph configuration.
+
+# List of Ceph disks.
```
markgoddard, line 92: Need the improved comment here.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+  become: True
+  when: ceph_disks | length > 0
+
+- name: Check the presence of a partition on the OSD disks
+  become: True
+  parted:
+    device: "{{ item.osd }}"
+  with_items: "{{ ceph_disks }}"
+  register: "disk_osd_info"
+  when: ceph_disks | length > 0
```
markgoddard: Not required here - if the list is empty there will be no iterations.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+  become: True
+  parted:
+    device: "{{ item.journal }}"
+  with_items: "{{ ceph_disks }}"
+  register: "disk_journal_info"
+  when: 
+    - ceph_disks | length > 0
+    - item.journal is defined
+
+- name: Test if the Ceph OSD disks have already a partition
```
markgoddard: I don't understand what this debug task is for. The user is unlikely to see this output unless we stop execution of the playbook.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+- name: Create tag partition for Ceph OSD
+  become: True
+  parted:
+    device: "{{ item.osd }}"
+    number: 1
+    label: gpt
+    name: "{% if item.journal is defined %} 'KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.osd | basename }}' {% else %} 'KOLLA_CEPH_OSD_BOOTSTRAP' {% endif %}"
+    state: present
+  with_items: "{{ ceph_disks }}"
+  when: not ceph_osd_test|skipped
```
markgoddard: I think we should iterate over disk_osd_info.results in this task, and apply the 'when' conditions here instead. There's no need for the debug task.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+  parted:
+    device: "{{ item.journal }}"
+    number: 1
+    label: gpt
+    name: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.osd | basename }}_J"
+    state: present
+  with_items: "{{ ceph_disks }}"
+  when:
+    - item.journal is defined
+    - ceph_journal_test|skipped
```
markgoddard: As above.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+    - item.journal is defined
+
+- name: Test if the Ceph OSD disks have already a partition
+  debug:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_osd_info.results }}"
+  when:
+    - not item | skipped
```
markgoddard: This won't ever be skipped, will it? Only journals can be skipped.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+  debug:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_journal_info.results }}"
+  when:
+    - not item | skipped
+    - item.partitions | length > 0
+    - not item.partitions.0.name.startswith('KOLLA_CEPH_OSD_BOOTSTRAP')
+    - ceph_disks | length > 0
```
markgoddard: ceph_disks | length > 0 not required.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,79 @@
...
+
+- name: Test if the Ceph journal disks have already a partition
+  debug:
+    msg: >
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_journal_info.results }}"
+  when:
+    - not item | skipped
+    - item.partitions | length > 0
```
markgoddard, line 47: Disks without partitions will be ignored.

markgoddard (changes requested):

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: you want a fail task ?

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: if disk list is total empty ?

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: yes, if disk have 0 partition it's good. so I skip it.

I use ceph_disks for tagging

ansible/roles/kolla-ceph/tasks/config.yml
markgoddard: If the list is emtpy, there will be no items.

ansible/roles/kolla-ceph/tasks/config.yml
markgoddard: If it has zero partitions, then you need to add a partition. But you won't because the parted task has when: not ceph_osd_test | skipped

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: So if I add test here and if disk have other partition, the tagging in skip but kayobe return just OK and SKIP instead of fail ?

kayobe/tests/unit/cli/test_commands.py
```diff
@@ -199,6 +199,7 @@ def test_seed_host_configure(self, mock_kolla_run, mock_run, mock_dump):
                     "ansible/kolla-target-venv.yml",
                     "ansible/kolla-host.yml",
                     "ansible/docker.yml",
+                    "ansible/ceph-block-devices.yml",
```
markgoddard: This (correctly) hasn't been added to seed host configure, which is why the test is failing. You can run locally via `tox -e py27`.

doc/source/release-notes.rst
```diff
@@ -43,6 +43,9 @@ Features
   this variable is ``{{ virtualenv_path }}/kolla-ansible``.
 * Adds tags to plays to support more fine grained configuration using the
```
markgoddard, line 44: pep8 is complaining about trailing whitespace in this file. You can run locally via `tox -e pep8`.

ansible/roles/kolla-ceph/tests/test-defaults.yml
```diff
@@ -0,0 +1,54 @@
...
+                 - osd: "{{ loopback_result.stdout }}"
+
+        - name: Get name of fake partition
+          parted_kayobe:
+            device: "{{ loopback_result.stdout }}"
+          register: "disk_osd_info"
+
+        - name: Validate OSD tag is present
+          assert:
+            that: "KOLLA_CEPH_OSD_BOOTSTRAP" in disk_osd_info.results.partitions.0.name
```
markgoddard: The travis job shows a syntax error here. Should this be (no .results without a with_*):

```
assert:
  that: "disk_osd_info.partitions.0.name == 'KOLLA_CEPH_OSD_BOOTSTRAP'"
```
You should be able to run the job locally with `tox -e ansible`.

markgoddard: You could also (first) assert that `disk_osd_info.partitions | length == 1`

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: Check the presence of a partition on the journal disks
+  become: True
+  parted_kayobe:
+    device: "{{ item.journal }}"
+  with_items: "{{ ceph_disks }}"
+  register: "disk_journal_info"
+  when: 
+    - item.journal is defined
+
+- name: Test if the Ceph OSD disks have already a partition
```
markgoddard: s/Test/Fail/

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+      The physical disk {{ item.item }} already has a partition.
+      Ensure that each disk in 'ceph_disks' does not have any partitions.
+  with_items: "{{ disk_osd_info.results }}"
+  when:
+    - item.partitions | length > 0
+    - not item.partitions.0.name.startswith('KOLLA_CEPH')
+  loop_control:
+    label: "{{item.item}}"
+
+- name: Test if the Ceph journal disks have already a partition
```
markgoddard: As above

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
+---
+# (ktibi) Need to remove parted_kayobe module when kayobe will support ansible 2.4
```
markgoddard: Also, remember we can't have it in the kayobe repo because it's GPL. I've created a repo: https://github.com/stackhpc/ansible-role-parted-1.1. Can you setup a PR into that repo with the parted module?

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: Create tag partition for Ceph OSD
+  become: True
+  parted_kayobe:
+    device: "{{ item.item.osd }}"
+    number: 1
+    label: gpt
+    name: "{{ part_label }}"
+    state: present
+  with_items: "{{ disk_osd_info.results }}"
+  when: item.partitions == 0
```
markgoddard: Should be item.partitions | length == 0

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+    name: "{{ part_label }}"
+    state: present
+  with_items: "{{ disk_journal_info.results }}"
+  when:
+    - not item | skipped
+    - item.partitions | length == 0
+  loop_control:
+    label: "{{item.item}}"
+  vars:
+      part_label: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.item.osd | basename }}{{ ansible_hostname }}_J"
```
markgoddard: Should be:
`part_label: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ part_id }}_J"`

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+    label: gpt
+    name: "{{ part_label }}"
+    state: present
+  with_items: "{{ disk_osd_info.results }}"
+  when: item.partitions == 0
+  loop_control:
+    label: "{{item.item}}"
+  vars:
+      part_label: "{% if item.item.journal is defined %}{{ part_label_with_journal }}{% else %}KOLLA_CEPH_OSD_BOOTSTRAP{% endif %}"
+      part_label_with_journal: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ part_label_id }}"
```
markgoddard: Should be:
`part_label_with_journal: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ part_id }}"`

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+    state: present
+  with_items: "{{ disk_journal_info.results }}"
+  when:
+    - not item | skipped
+    - item.partitions | length == 0
+  loop_control:
+    label: "{{item.item}}"
+  vars:
+      part_label: "KOLLA_CEPH_OSD_BOOTSTRAP_{{ item.item.osd | basename }}{{ ansible_hostname }}_J"
+      part_id: "{{ item.item.osd | basename }}{{ ansible_hostname }}"
```
markgoddard: If there is a better name for this variable, feel free to change it.

ansible/roles/kolla-ceph/tasks/config.yml
```diff
@@ -0,0 +1,86 @@
...
+- name: Check the presence of a partition on the OSD disks
+  become: True
+  parted_kayobe:
+    device: "{{ item.osd }}"
+  with_items: "{{ ceph_disks }}"
+  register: "disk_osd_info"
+
+- name: Check the presence of a partition on the journal disks
+  become: True
+  parted_kayobe:
```
markgoddard: Maybe name this parted_1_1?

markgoddard (changes requested):

ansible/roles/kolla-ceph/tasks/config.yml
ktibi: yes it's not a partition but the ID of OSD.

ansible/roles/kolla-ceph/tests/test-defaults.yml
```diff
@@ -0,0 +1,60 @@
...
+    - block:
+        - name: Test the kolla-ceph role
+          include_role:
+            name:  ../../kolla-ceph
+          vars:
+              ceph_disks:
+                 - osd: "{{ loopback_result.stdout }}"
+
+        - name: Get name of fake partition
+          parted_kayobe:
```
markgoddard: Still using parted_kayobe here.

markgoddard (approved):
Tests are now passing.

PR #85, opened 2017-11-28
[Vagrant] Support for VMware Fusion, fixes for container image builds

yankcrime:
This PR:

* Refactors the Vagrantfile somewhat and introduces support for VMware Fusion;
* In order to support HGFS for the above, it also switches the default Vagrant boxes to be ones hosted under the [StackHPC organisation](https://app.vagrantup.com/stackhpc/boxes/centos-7) on Vagrant Cloud.  These have been built with the relevant guest additions which are missing from the default CentOS images due to licensing shenanigans;
*  Adds the development VM to a `container-image-builders` group so that images can be built.

### DIFF
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,82 +1,26 @@
 # -*- mode: ruby -*-
 # vi: set ft=ruby :
 
-# All Vagrant configuration is done below. The "2" in Vagrant.configure
-# configures the configuration version (we support older styles for
-# backwards compatibility). Please don't change it unless you know what
-# you're doing.
-Vagrant.configure("2") do |config|
-  # The most common configuration options are documented and commented below.
-  # For a complete reference, please see the online documentation at
-  # https://docs.vagrantup.com.
+Vagrant.configure('2') do |config|
+  config.vm.hostname = 'controller1'
 
-  # Every Vagrant development environment requires a box. You can search for
-  # boxes at https://atlas.hashicorp.com/search.
-  config.vm.box = "centos/7"
+  config.vm.network 'private_network', ip: '192.168.33.3', auto_config: false
 
-  # Disable automatic box update checking. If you disable this, then
-  # boxes will only be checked for updates when the user runs
-  # `vagrant box outdated`. This is not recommended.
-  # config.vm.box_check_update = false
+  config.vm.box = 'stackhpc/centos-7'
 
-  # The default hostname is localhost, so set it to something a little more
-  # sensible.
-  config.vm.hostname = "controller1"
-
-  # Create a forwarded port mapping which allows access to a specific port
-  # within the machine from a port on the host machine. In the example below,
-  # accessing "localhost:8080" will access port 80 on the guest machine.
-  # config.vm.network "forwarded_port", guest: 80, host: 8080
-
-  # Create a private network, which allows host-only access to the machine
-  # using a specific IP.
-  config.vm.network "private_network", ip: "192.168.33.3", auto_config: false
-
-  # Create a public network, which generally matched to bridged network.
-  # Bridged networks make the machine appear as another physical device on
-  # your network.
-  # config.vm.network "public_network"
-
-  # Share an additional folder to the guest VM. The first argument is
-  # the path on the host to the actual folder. The second argument is
-  # the path on the guest to mount the folder. And the optional third
-  # argument is a set of non-required options.
-
-  # NOTE: To make this work install vbguest plugin to install tools in VM:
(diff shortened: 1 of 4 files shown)


Vagrantfile
```diff
@@ -1,82 +1,26 @@
...
-  # Every Vagrant development environment requires a box. You can search for
-  # boxes at https://atlas.hashicorp.com/search.
-  config.vm.box = "centos/7"
+  config.vm.network 'private_network', ip: '192.168.33.3', auto_config: false
 
-  # Disable automatic box update checking. If you disable this, then
-  # boxes will only be checked for updates when the user runs
-  # `vagrant box outdated`. This is not recommended.
-  # config.vm.box_check_update = false
+  config.vm.box = 'stackhpc/centos-7'
```
markgoddard, line 9: How did you build this box? Is there a Vagrantfile we can add for if we ever need to rebuild it?

Vagrantfile
```diff
@@ -1,82 +1,26 @@
...
-  # config.vm.network "public_network"
-
-  # Share an additional folder to the guest VM. The first argument is
-  # the path on the host to the actual folder. The second argument is
-  # the path on the guest to mount the folder. And the optional third
-  # argument is a set of non-required options.
-
-  # NOTE: To make this work install vbguest plugin to install tools in VM:
-  #   vagrant plugin install vagrant-vbguest
-  config.vm.synced_folder ".", "/vagrant", type: "virtualbox"
```
markgoddard, line 47: We found that we need this, or else we don't get a persistent share. Can you add it inside the virtualbox provider?

dev/dev-hosts
```diff
@@ -1,2 +1,5 @@
 [controllers]
-controller0 ansible_host=192.168.33.3
+controller1 ansible_host=192.168.33.3
+
+[container-image-builders]
+controller1 ansible_host=192.168.33.3
```
markgoddard, line 5: There's also one in dev/dev-vagrant-network-allocation.yml which is controller0. I think the Vagrantfile was probably the odd one out, but it doesn't matter too much which way we go.

markgoddard (changes requested):
Looks good, thanks.

Vagrantfile
yankcrime: They're generated via Packer using a fork of Boxcutter's [CentOS templates](https://github.com/boxcutter/centos).  I'll move the fork to sit under StackHPC's organisation so that we've all got visibility of how these are built.

Vagrantfile
yankcrime: You'll get a persistent share using this updated image - I believe it's supposed to be the default as long as Vagrant can detect presence of the VirtualBox guest tools in the VM as it's booting, otherwise it falls back to rsync.  Happy to add this back in though so it's explicit 👍

markgoddard: As long as it works, I don't mind!

Vagrantfile
markgoddard: :+1:

markgoddard:
Are you able to test this with virtualbox, or shall I do it?

yankcrime:
I've tested it on a machine I have here running Ubuntu 17.10 and VirtualBox 5.1 but it'd be great if you could give it a go as well - especially if you're running different versions to me, just to make sure the guest additions are detected OK and that shared folders Just Work ™:

```
==> default: Machine booted and ready!
[default] GuestAdditions 5.1.30 running --- OK.
==> default: Checking for guest additions in VM...
==> default: Setting hostname...
==> default: Mounting shared folders...
    default: /vagrant => /home/nick/src/kayobe
```

dev/dev-hosts
yankcrime: Done!

Vagrantfile
markgoddard: It worked for me without, but then looking at the commit history, JohnG added it. It's possible that use of the vbguest plugin means that it will now always work. Perhaps @JohnGarbutt could verify that it works for him if we're going to remove it?

markgoddard (approved):
I've deployed a VM on my laptop with these changes and all worked well.

There's still a question around the vagrant shared folder for JohnG to answer but otherwise I'm happy.

Vagrantfile
markgoddard: Here's the commit: https://github.com/stackhpc/kayobe/commit/42605939a4f911fb3bf777e2d80e91cf3f5aa710

markgoddard:
I don't see JohnG replying any time soon, so let's merge this.

PR #86, opened 2017-11-28
add collectd regex

ktibi:


markgoddard (changes requested):
Do you also need to add a regex for the container image here: https://github.com/stackhpc/kayobe/blob/master/ansible/group_vars/all/kolla#L106? That would also require adding a default for `kolla_enable_collectd` in the same file.

ktibi:
done

markgoddard (approved):
Thanks

PR #89, opened 2017-12-04
Added python-ironicclient

autonomouse:
(could also go in the os-openstackclient role, but that's on galaxy)


markgoddard (approved):
Looks good. Added a blank line after the task for consistency.

PR #97, opened 2017-12-13
Add support for virtual environments for remote execution

markgoddard:


ansible/roles/kolla-ansible/templates/requirements.txt.j2
```diff
@@ -0,0 +1,8 @@
+{% if kolla_ansible_ctl_install_type == 'source' %}
+# Install Kolla Ansible from source.
+{{ kolla_ansible_source_path }}
+{% else %}
+# Install Kolla Ansible from PyPI.
+kolla-ansible=={{ kolla_openstack_release }}
+{% endif %}
+ansible<2.4
```
dougszumski: nit: For future reference it would be nice to have a comment explaining why we restrict ourselves to <2.4

dougszumski (approved):
Looks good to me, but I didn't test.

dougszumski (approved):
:+1:

PR #105, opened 2018-01-03
Fix storage config for docker-ce

ktibi:
New version of docker-ce use overlay for storage driver by default. 

We need to adapt the check for test if overlay is enable and if devicemapper is configured in kayobe.

### DIFF
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -85,8 +85,8 @@
     state: started
   become: True
 
-- name: Check whether docker storage is in loopback mode
-  command: docker info
+- name: Query docker daemon information
+  command: "docker info"
   register: docker_info
   changed_when: False
 
@@ -96,11 +96,10 @@
       Not configuring docker storage in {{ docker_storage_driver }} mode as
       loopback-backed containers or images exist.
   when:
-    - "'Data loop file' in docker_info.stdout"
-    - "'Images: 0' not in docker_info.stdout"
-    - "'Containers: 0' not in docker_info.stdout"
+    - "'Data loop file' in docker_info.stdout or docker_storage_driver not in docker_info.stdout"
+    - "'Images: 0' not in docker_info.stdout or 'Containers: 0' not in docker_info.stdout"
 
 - include: storage.yml
-  when: "'Data loop file' in docker_info.stdout"
+  when: "'Data loop file' in docker_info.stdout or docker_storage_driver not in docker_info.stdout"
 
 - include: config.yml


ansible/roles/docker/tasks/main.yml
```diff
@@ -101,6 +101,6 @@
     - "'Containers: 0' not in docker_info.stdout"
 
 - include: storage.yml
-  when: "'Data loop file' in docker_info.stdout"
+  when: "'Data loop file' in docker_info.stdout or ( docker_storage_driver == 'devicemapper' and 'Storage Driver: overlay' in docker_info.stdout )"
```
markgoddard: This needs to check if the actual storage driver == expected storage driver, not just the specific case of overlay -> devicemapper.

ansible/roles/docker/tasks/main.yml
```diff
@@ -101,6 +101,6 @@
     - "'Containers: 0' not in docker_info.stdout"
```
markgoddard: We need to add the check in this task (L99) as well - we can't change the storage driver if images or containers exist.

markgoddard: Optional: To make parsing the output easier, we could decode the docker info output from JSON:

```
 - name: Query docker daemon information
   command: "docker info -f {{ json . }}"
   register: docker_info_json
   changed_when: False

- name: Decode the docker info output
  set_fact:
    docker_info: "{{ docker_info_json | from_json }}"
```

Now the storage driver is accessible via `{{ docker_info.Driver }}`. I'm not sure what the name of the 'Data loop file' field would be in JSON format though, perhaps something under `{{ docker_info.DriverStatus }}`.

markgoddard (changes requested):

ansible/roles/docker/tasks/main.yml
```diff
@@ -85,22 +85,25 @@
...
 - name: Fail when loopback-mode containers or images exist
   fail:
     msg: >
       Not configuring docker storage in {{ docker_storage_driver }} mode as
       loopback-backed containers or images exist.
   when:
-    - "'Data loop file' in docker_info.stdout"
-    - "'Images: 0' not in docker_info.stdout"
-    - "'Containers: 0' not in docker_info.stdout"
+    - "'Data loop file' in docker_info.DriverStatus or docker_storage_driver != docker_info.driver"
```
markgoddard: docker_info.Driver is capitalised for me.

Also, DriverStatus is a list of lists, so the first expression (data loop file) won't ever pass.

markgoddard (changes requested):

ansible/roles/docker/tasks/main.yml
```diff
@@ -96,11 +96,10 @@
...
   when:
-    - "'Data loop file' in docker_info.stdout"
-    - "'Images: 0' not in docker_info.stdout"
-    - "'Containers: 0' not in docker_info.stdout"
+    - "'Data loop file' in docker_info.stdout or docker_storage_driver not in docker_info.stdout"
+    - "'Images: 0' not in docker_info.stdout or 'Containers: 0' not in docker_info.stdout"
 
 - include: storage.yml
-  when: "'Data loop file' in docker_info.stdout"
+  when: "'Data loop file' in docker_info.stdout or docker_storage_driver not int docker_info.stdout"
```
markgoddard: int

ktibi: done

markgoddard (approved):
Looks fine, assuming it's been tested?

ktibi:
I'll test last version tomorrow. Wait before merge.

ktibi:
@markgoddard Works ! can be merged !

markgoddard (approved):

PR #106, opened 2018-01-08
Fix ipa image copy

diranetafen:
Hi
When we build the IPA images we must always copy them in bifrost otherwise they will not copy if they have already been built in the past. We meet this case when redeploying the seed, the images are not present in / httpboot /.
So we propose to systematically copy the images even if


ansible/seed-ipa-build.yml
```diff
@@ -35,18 +35,15 @@
...
-      with_items: "{{ ipa_images }}"
+        - name: Copy Ironic Python Agent images into /httpboot
+          command: >
+            docker exec bifrost_deploy
+            bash -c 'source /bifrost/env-vars &&
+            ansible -vvvv target -i /bifrost/playbooks/inventory/target
+            -m copy
+            -a "src=/etc/bifrost/{{ item }} dest=/httpboot/{{ item }}"
+            -e "ansible_python_interpreter=/var/lib/kolla/venv/bin/python"'
+          with_items: "{{ ipa_images }}"
```
markgoddard, line 48: This syntax is incorrect. The task must come before the 'when'. Has it been tested?

markgoddard (changes requested):

ansible/seed-ipa-build.yml
diranetafen: Hi Mark, I've corrected my pull request (#107 ) and I will close this one.

markgoddard: Thanks

PR #107, opened 2018-01-08
Fix copy image ipa on seed

diranetafen:
Hi
When we build the IPA images we must always copy them in bifrost otherwise they will not copy if they have already been built in the past. We meet this case when redeploying the seed, the images are not present in / httpboot /.
So we propose to systematically copy the images even if


markgoddard (approved):
Thanks

PR #108, opened 2018-01-11
Add extra conf for cinder and ceph

ktibi:


ansible/kolla-openstack.yml
```diff
@@ -192,6 +192,8 @@
       kolla_inspector_dhcp_pool_end: "{{ inspection_net_name | net_inspection_allocation_pool_end }}"
       kolla_inspector_default_gateway: "{{ inspection_net_name | net_inspection_gateway or inspection_net_name | net_gateway }}"
       # Extra free-form user-provided configuration.
+      kolla_extra_ceph: "{{ kolla_extra_config.ceph | default }}"
+      kolla_extra_cinder: "{{ kolla_extra_config.ciner | default }}"
```
markgoddard: s/ciner/cinder/

markgoddard: Also need to add these on L99

markgoddard (changes requested):

ansible/kolla-openstack.yml
ktibi: done

markgoddard (approved):
Looks better. Tested?

ktibi:
yes tested and work

PR #109, opened 2018-01-12
Add support for configuring Grafana

dougszumski:


ansible/roles/kolla-openstack/defaults/main.yml
```diff
@@ -15,6 +15,15 @@ kolla_enable_glance:
 # glance-registry.conf.
 kolla_extra_glance:
 
+###############################################################################
+# Grafana configuration.
+
+# Whether to enable Grafana.
+kolla_enable_grafana:
```
markgoddard, line 22: There needs to be a default for this here: https://github.com/stackhpc/kayobe/blob/master/ansible/group_vars/all/kolla#L272. Might be best to check what the default is for enable_grafana in kolla-ansible.

markgoddard (changes requested):
Config is mostly fine, but we need to add grafana to the image regex map as well to ensure it gets built when necessary.

ansible/roles/kolla-openstack/defaults/main.yml
dougszumski: Thanks - ready for re-review.

markgoddard (approved):

PR #114, opened 2018-01-17
Support configuring API interface via overcloud inventory file

dougszumski:
I've probably missed something. Once this is good, I can apply it to the other interfaces. It solves my issue with deploying grafana on a multinode host at least.

### DIFF
--- a/ansible/kolla-ansible.yml
+++ b/ansible/kolla-ansible.yml
@@ -10,6 +10,19 @@
       set_fact:
         kolla_api_interface: "{{ internal_net_name | net_interface | replace('-', '_')  }}"
       when: internal_net_name in network_interfaces
+    - name: Validate overcloud host Kolla Ansible network configuration
+      fail:
+        msg: >
+          The Kolla Ansible variable {{ item.var_name }}
+          ({{ item.description }}) is invalid. Value:
+          "{{ hostvars[inventory_hostname][item.var_name] | default('<undefined>') }}".
+      when:
+        - item.required | bool
+        - hostvars[inventory_hostname][item.var_name] is not defined or not hostvars[inventory_hostname][item.var_name]
+      with_items:
+        - var_name: "kolla_api_interface"
+          description: "API network interface name"
+          required: True
 
 - name: Ensure Kolla Ansible is configured
   hosts: localhost
@@ -63,9 +76,6 @@
             - item.required | bool
             - hostvars[inventory_hostname][item.var_name] is not defined or not hostvars[inventory_hostname][item.var_name]
           with_items:
-            - var_name: "kolla_api_interface"
-              description: "API network interface name"
-              required: True
             - var_name: "kolla_provision_interface"
               description: "Bare metal provisioning network interface name"
               required: "{{ kolla_enable_ironic }}"
--- a/ansible/kolla-ansible.yml
+++ b/ansible/kolla-ansible.yml
@@ -1,4 +1,16 @@
 ---
+- name: Set interfaces for overcloud hosts
+  hosts: overcloud
+  tags:
+    - config
+    - kolla-ansible
+  gather_facts: False
+  tasks:
+    - name: Set API interface
+      set_fact:
+        kolla_api_interface: "{{ internal_net_name | net_interface | replace('-', '_')  }}"
+      when: internal_net_name in network_interfaces
+
 - name: Ensure Kolla Ansible is configured
   hosts: localhost
   tags:
@@ -35,7 +47,6 @@
         - name: Set facts containing the controller network interfaces
           set_fact:
             kolla_network_interface: "{{ internal_net_name | net_interface(controller_host) | replace('-', '_') }}"
-            kolla_api_interface: "{{ internal_net_name | net_interface(controller_host) | replace('-', '_') }}"
             kolla_storage_interface: "{{ storage_net_name | net_interface(controller_host) | replace('-', '_') }}"
             kolla_cluster_interface: "{{ storage_mgmt_net_name | net_interface(controller_host) | replace('-', '_') }}"
             kolla_provision_interface: "{{ provision_wl_net_name | net_interface(controller_host) | replace('-', '_') }}"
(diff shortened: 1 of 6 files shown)


dougszumski:
Example overcloud inventory generated with this patch:

```
# Top level controllers group.
[controllers]
# These hostnames must be resolvable from your deployment host
s1 ansible_host=10.0.0.101 api_interface=breno1.7

# Top level compute group.
[compute]
# These hostnames must be resolvable from your deployment host

# Top level monitoring group.
[monitoring]
# These hostnames must be resolvable from your deployment host
s2 ansible_host=10.0.0.102 api_interface=eno1.7

# Top level network group.
[network]
# These hostnames must be resolvable from your deployment host
s1 ansible_host=10.0.0.101 api_interface=breno1.7


```

ansible/kolla-ansible.yml
```diff
@@ -1,4 +1,15 @@
 ---
+- name: Set interfaces for overcloud hosts
+  hosts: overcloud
+  tags:
+    - config
```
markgoddard, line 5: Also needs kolla-ansible tag.

ansible/roles/kolla-ansible/defaults/main.yml
```diff
@@ -69,6 +69,7 @@ kolla_overcloud_inventory_pass_through_host_vars:
   - "ansible_host"
   - "ansible_port"
   - "ansible_ssh_private_key_file"
+  - "kolla_api_interface"
```
markgoddard, line 72: Interesting, hadn't thought to use this.

ansible/roles/kolla-ansible/templates/overcloud-top-level.j2
```diff
@@ -17,7 +17,7 @@
 # These hostnames must be resolvable from your deployment host
 {% for host in groups.get(group, []) %}
 {% set host_hv=hostvars[host] %}
-{{ host }}{% for hv_name in kolla_overcloud_inventory_pass_through_host_vars %}{% if hv_name in host_hv %} {{ hv_name }}={{ host_hv[hv_name] }}{% endif %}{% endfor %}
+{{ host }}{% for hv_name in kolla_overcloud_inventory_pass_through_host_vars %}{% if hv_name in host_hv %} {{ hv_name | regex_replace('^kolla_(.*)$', '\1') }}={{ host_hv[hv_name] }}{% endif %}{% endfor %}
```
markgoddard, line 20: I think once all interfaces use this method our inventory lines are going to be a little silly. That said, if this gets you moving again then I'm fine with it for now. After all, what's one more wafer thin mint?

markgoddard: I don't think there are any variables that we might want to pass through that should keep their kolla_ prefix, so I guess that's not a problem.

markgoddard (changes requested):
Looks good except for the tag.

ansible/roles/kolla-ansible/templates/overcloud-top-level.j2
dougszumski: Agreed. ```api_interface``` is the only one I need at the moment, so I don't have to convert the others if you'd rather leave those for now.

markgoddard: I think there are two options I'd be happy with.
1. Stick with this as a means to an end
2. Support all interfaces and write the variables to an `inventory/host_vars/<host>` file. Also in this case hard code the interface variable names rather than adding them to the pass through vars list.

markgoddard:
Ah, also the ansible unit tests are failing!

ansible/roles/kolla-ansible/templates/overcloud-top-level.j2
dougszumski: Thanks - in the interest of making progress I'll go with 1 for now and create an issue for 2.

dougszumski:
Tests passed - rebased for merge.

Issue created here for other interfaces: https://github.com/stackhpc/kayobe/issues/115

markgoddard (approved):

dougszumski:
Sorry - one extra commit.

ansible/kolla-ansible.yml
```diff
@@ -52,9 +63,6 @@
             - item.required | bool
             - hostvars[inventory_hostname][item.var_name] is not defined or not hostvars[inventory_hostname][item.var_name]
           with_items:
-            - var_name: "kolla_api_interface"
-              description: "API network interface name"
-              required: True
```
markgoddard, line 57: We do need to verify this. Could you add it to the new play above instead?

markgoddard (changes requested):

ansible/kolla-ansible.yml
dougszumski: Thanks - done.

ansible/roles/kolla-ansible/templates/overcloud-top-level.j2
markgoddard: Thinking about this again, I think we're going to have issues with the seed, which currently needs to have an api_interface set in kolla-ansible, even though it's not used: https://github.com/openstack/kolla-ansible/blob/master/ansible/roles/baremetal/tasks/pre-install.yml#L19.

markgoddard (changes requested):

ansible/roles/kolla-ansible/templates/overcloud-top-level.j2
markgoddard: Actually, it might just work as it is. api_interface for the seed will default to network_interface, which is set to the same value that api_interface was previously in globals.yml.

markgoddard (approved):

ansible/roles/kolla-ansible/templates/overcloud-top-level.j2
dougszumski: I was just having a dig around and wondering if I needed to explicitly set it - thanks for the pointer.

PR #116, opened 2018-01-19
Support configuring Grafana

dougszumski:
This add supports for running the stackhpc.grafana-conf Ansible
Galaxy role against a monitoring server to configure Grafana
post deployment. This includes:

* Creating an organisation
* Configuring dashboards for the organisation from a git repo
* Configuring datasources for the organisation

### DIFF
--- a/ansible/group_vars/all/grafana
+++ b/ansible/group_vars/all/grafana
@@ -1,0 +1,50 @@
+---
+###############################################################################
+# Grafana configuration.
+
+# Grafana local admin user name. If you are deploying Monasca Grafana this
+# should not conflict with an OpenStack user name.
+#grafana_local_admin_user_name:
+grafana_local_admin_user_name: "grafana_admin"
+
+# Path to git repo containing Grafana dashboards. Eg.
+# https://github.com/stackhpc/grafana-reference-dashboards.git
+#grafana_monitoring_node_dashboard_repo:
+grafana_monitoring_node_dashboard_repo:
+
+# Path to which Grafana dashboards will be cloned to a monitoring node
+#grafana_monitoring_node_dashboard_repo_path:
+grafana_monitoring_node_dashboard_repo_path: "/var/opt/grafana/"
+
+# The Grafana organisation for the control plane. Note that for Monasca
+# Grafana with domain support the format is:
+# organisation_name@openstack_domain
+#grafana_control_plane_organisation:
+grafana_control_plane_organisation: "control_plane"
+
+# A list of datasources to configure. See the stackhpc.grafana-conf role
+# for a list of supported datasources. Example:
+#
+# grafana_datasources:
+#   monasca_api:
+#     port: 8082
+#     host: monasca-api
+#   monasca_log_api:
+#     port: 5607
+#     host: monasca-log-api
+#   elasticsearch:
+#     port: 9200
+#     host: monasca-elasticsearch
+#     project_id: "some_id"
+#
+#grafana_datasources:
+grafana_datasources:
+
+# The path, relative to the grafana_monitoring_node_dashboard_repo_path
+# containing the dashboards. Eg. /prometheus/control_plane
+#grafana_monitoring_node_dashboard_path:
+grafana_monitoring_node_dashboard_path:
+
+###############################################################################
+# Dummy variable to allow Ansible to accept this file.
+workaround_ansible_issue_8743: yes
--- a/ansible/group_vars/all/grafana
+++ b/ansible/group_vars/all/grafana
@@ -25,8 +25,8 @@ grafana_monitoring_node_dashboard_repo_path:
 # organisation_name@openstack_domain
 grafana_control_plane_organisation: "control_plane"
 
-# A list of datasources to configure. See the stackhpc.grafana-conf role
-# for a list of supported datasources. Example:
+# A dict of datasources to configure. See the stackhpc.grafana-conf role
+# for all supported datasources. Example:
 #
 # grafana_datasources:
 #   monasca_api:
--- a/ansible/group_vars/all/grafana
+++ b/ansible/group_vars/all/grafana
@@ -10,8 +10,15 @@ grafana_local_admin_user_name: "grafana_admin"
 # https://github.com/stackhpc/grafana-reference-dashboards.git
 grafana_monitoring_node_dashboard_repo:
 
+# Dashboard repo version. Optional, defaults to 'HEAD'.
+grafana_monitoring_node_dashboard_repo_version:
+
 # Path to which Grafana dashboards will be cloned to a monitoring node
-grafana_monitoring_node_dashboard_repo_path: "{{ source_checkout_path ~ '/grafana' }}"
+grafana_monitoring_node_dashboard_repo_checkout_path: "{{ source_checkout_path ~ '/grafana-dashboards' }}"
+
+# The path, relative to the grafana_monitoring_node_dashboard_repo_checkout_path
+# containing the dashboards. Eg. /prometheus/control_plane
+grafana_monitoring_node_dashboard_repo_path:
 
 # The Grafana organisation for the control plane. Note that for Monasca
 # Grafana with domain support the format is:
@@ -34,9 +41,5 @@ grafana_control_plane_organisation: "control_plane"
 #     project_id: "some_id"
 #
 grafana_datasources: {}
-
-# The path, relative to the grafana_monitoring_node_dashboard_repo_path
-# containing the dashboards. Eg. /prometheus/control_plane
-grafana_monitoring_node_dashboard_path:
 
 ###############################################################################
--- a/ansible/group_vars/all/grafana
+++ b/ansible/group_vars/all/grafana
@@ -4,22 +4,18 @@
 
 # Grafana local admin user name. If you are deploying Monasca Grafana this
 # should not conflict with an OpenStack user name.
-#grafana_local_admin_user_name:
 grafana_local_admin_user_name: "grafana_admin"
 
 # Path to git repo containing Grafana dashboards. Eg.
 # https://github.com/stackhpc/grafana-reference-dashboards.git
-#grafana_monitoring_node_dashboard_repo:
 grafana_monitoring_node_dashboard_repo:
 
 # Path to which Grafana dashboards will be cloned to a monitoring node
-#grafana_monitoring_node_dashboard_repo_path:
-grafana_monitoring_node_dashboard_repo_path: "/var/opt/grafana/"
+grafana_monitoring_node_dashboard_repo_path: "{{ source_checkout_path ~ '/grafana' }}"
 
 # The Grafana organisation for the control plane. Note that for Monasca
 # Grafana with domain support the format is:
 # organisation_name@openstack_domain
-#grafana_control_plane_organisation:
 grafana_control_plane_organisation: "control_plane"
 
 # A list of datasources to configure. See the stackhpc.grafana-conf role
@@ -37,14 +33,10 @@ grafana_control_plane_organisation: "control_plane"
 #     host: monasca-elasticsearch
 #     project_id: "some_id"
 #
-#grafana_datasources:
-grafana_datasources:
+grafana_datasources: {}
 
 # The path, relative to the grafana_monitoring_node_dashboard_repo_path
 # containing the dashboards. Eg. /prometheus/control_plane
-#grafana_monitoring_node_dashboard_path:
 grafana_monitoring_node_dashboard_path:
 
 ###############################################################################
-# Dummy variable to allow Ansible to accept this file.
-workaround_ansible_issue_8743: yes
--- a/ansible/group_vars/all/grafana
+++ b/ansible/group_vars/all/grafana
@@ -4,7 +4,7 @@
 
 # Grafana local admin user name. If you are deploying Monasca Grafana this
 # should not conflict with an OpenStack user name.
-grafana_local_admin_user_name: "grafana_admin"
+grafana_local_admin_user_name: "admin"
 
 # Path to git repo containing Grafana dashboards. Eg.
 # https://github.com/stackhpc/grafana-reference-dashboards.git
--- a/ansible/overcloud-grafana-configure.yml
+++ b/ansible/overcloud-grafana-configure.yml
@@ -1,0 +1,44 @@
+---
+- name: Check whether Grafana is enabled
+  hosts: overcloud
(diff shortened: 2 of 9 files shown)


ansible/overcloud-grafana-configure.yml
```diff
@@ -0,0 +1,44 @@
...
+      set_fact:
+        kolla_internal_vip_address: "{{ internal_net_name | net_vip_address }}"
+      when: kolla_enable_haproxy | bool
+
+    - name: Set fact for the VIP address
+      set_fact:
+        kolla_internal_vip_address: "{{ internal_net_name | net_ip(network_host) }}"
+      when: not kolla_enable_haproxy | bool
+
+    - name: Clone dashboards
```
dougszumski: I should probably make the Galaxy role do this, as at the very least it will also be required for using the role for configuring tenant monitoring.

kayobe/cli/commands.py
```diff
@@ -1087,14 +1087,15 @@ class OvercloudPostConfigure(KayobeAnsibleMixin, VaultMixin, Command):
...
     """
 
     def take_action(self, parsed_args):
         self.app.LOG.debug("Performing post-deployment configuration")
         playbooks = _build_playbook_list(
             "overcloud-ipa-images", "overcloud-introspection-rules",
             "overcloud-introspection-rules-dell-lldp-workaround",
-            "provision-net")
+            "provision-net", "overcloud-grafana-configure")
         self.run_kayobe_playbooks(parsed_args, playbooks)
```
markgoddard, line 1099: I've been in the habit of adding unit tests for commands as I change them. Would you mind adding one for overcloud post configure?

requirements.yml
```diff
@@ -10,6 +10,7 @@
   name: singleplatform-eng.users
 - src: stackhpc.drac
 - src: stackhpc.drac-facts
+- src: stackhpc.grafana-conf
```
markgoddard, line 13: Can you add the role to .gitignore?

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,50 @@
+---
+###############################################################################
+# Grafana configuration.
+
+# Grafana local admin user name. If you are deploying Monasca Grafana this
+# should not conflict with an OpenStack user name.
+#grafana_local_admin_user_name:
```
markgoddard: nit: I don't normally put the commented out variable in group_vars, only etc/kayobe/

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,50 @@
...
+grafana_local_admin_user_name: "grafana_admin"
+
+# Path to git repo containing Grafana dashboards. Eg.
+# https://github.com/stackhpc/grafana-reference-dashboards.git
+#grafana_monitoring_node_dashboard_repo:
+grafana_monitoring_node_dashboard_repo:
+
+# Path to which Grafana dashboards will be cloned to a monitoring node
+#grafana_monitoring_node_dashboard_repo_path:
+grafana_monitoring_node_dashboard_repo_path: "/var/opt/grafana/"
```
markgoddard: I tend to use `{{ source_checkout_path }}/grafana`.

ansible/overcloud-grafana-configure.yml
```diff
@@ -0,0 +1,44 @@
...
+      git:
+        repo: "{{ grafana_monitoring_node_dashboard_repo }}"
+        dest: "{{ grafana_monitoring_node_dashboard_repo_path }}"
+      when: grafana_monitoring_node_dashboard_repo | bool
+      become: True
```
markgoddard: Do we need become here? In similar situations I've used the file module with become to ensure the parent directory exists, then cloned the repo without.

ansible/overcloud-grafana-configure.yml
```diff
@@ -0,0 +1,44 @@
...
+
+    - name: Include Kolla passwords for Grafana local admin account credentials
+      include_vars: "{{ kayobe_config_path }}/kolla/passwords.yml"
+  roles:
+    - role: stackhpc.grafana-conf
+      grafana_conf_organisation: "{{ grafana_control_plane_organisation }}"
+      grafana_conf_grafana_admin_user: "{{ grafana_local_admin_user_name }}"
```
markgoddard, line 33: Should we set this in grafana.conf too?

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,50 @@
...
+grafana_datasources:
+
+# The path, relative to the grafana_monitoring_node_dashboard_repo_path
+# containing the dashboards. Eg. /prometheus/control_plane
+#grafana_monitoring_node_dashboard_path:
+grafana_monitoring_node_dashboard_path:
+
+###############################################################################
+# Dummy variable to allow Ansible to accept this file.
+workaround_ansible_issue_8743: yes
```
markgoddard: nit: Don't need this in group_vars as there is always at least one variable.

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,50 @@
...
+#   monasca_log_api:
+#     port: 5607
+#     host: monasca-log-api
+#   elasticsearch:
+#     port: 9200
+#     host: monasca-elasticsearch
+#     project_id: "some_id"
+#
+#grafana_datasources:
+grafana_datasources:
```
markgoddard: Should this default to `{}`?

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,50 @@
...
+
+# Grafana local admin user name. If you are deploying Monasca Grafana this
+# should not conflict with an OpenStack user name.
+#grafana_local_admin_user_name:
+grafana_local_admin_user_name: "grafana_admin"
+
+# Path to git repo containing Grafana dashboards. Eg.
+# https://github.com/stackhpc/grafana-reference-dashboards.git
+#grafana_monitoring_node_dashboard_repo:
+grafana_monitoring_node_dashboard_repo:
```
markgoddard: Might we ever need multiple dashboard repos?

requirements.yml
dougszumski: Done

kayobe/cli/commands.py
dougszumski: Good idea, i've added a basic test to ensure the correct playbooks are called.

ansible/overcloud-grafana-configure.yml
dougszumski: Good idea - I was setting it in the config repo, but it's better done in Kayobe.

ansible/overcloud-grafana-configure.yml
dougszumski: I've moved this to the ansible role and used file to ensure the dir exists first.

ansible/group_vars/all/grafana
dougszumski: Fixed

ansible/group_vars/all/grafana
dougszumski: Fixed

ansible/group_vars/all/grafana
dougszumski: Done

ansible/group_vars/all/grafana
dougszumski: Hmm, I can't see a use at the moment. I think it's better to have all the dashboard config in one repo for simplicity. I can always add it later if we need it.

ansible/group_vars/all/grafana
dougszumski: Done

ansible/overcloud-grafana-configure.yml
dougszumski: Done

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,42 @@
...
+
+# Path to which Grafana dashboards will be cloned to a monitoring node
+grafana_monitoring_node_dashboard_repo_path: "{{ source_checkout_path ~ '/grafana' }}"
+
+# The Grafana organisation for the control plane. Note that for Monasca
+# Grafana with domain support the format is:
+# organisation_name@openstack_domain
+grafana_control_plane_organisation: "control_plane"
+
+# A list of datasources to configure. See the stackhpc.grafana-conf role
```
markgoddard: nit: dict

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,42 @@
...
+#   elasticsearch:
+#     port: 9200
+#     host: monasca-elasticsearch
+#     project_id: "some_id"
+#
+grafana_datasources: {}
+
+# The path, relative to the grafana_monitoring_node_dashboard_repo_path
+# containing the dashboards. Eg. /prometheus/control_plane
+grafana_monitoring_node_dashboard_path:
```
markgoddard: Should this be with the other repo variables above?

ansible/overcloud-grafana-configure.yml
```diff
@@ -0,0 +1,40 @@
...
+    - grafana
+  pre_tasks:
+    - name: Set fact for the VIP address
+      set_fact:
+        kolla_internal_vip_address: "{{ internal_net_name | net_vip_address }}"
+      when: kolla_enable_haproxy | bool
+
+    - name: Set fact for the VIP address
+      set_fact:
+        kolla_internal_vip_address: "{{ internal_net_name | net_ip(network_host) }}"
```
markgoddard: I don't think network_host has been set. If there's no haproxy then I think we can assume that grafana is running on the monitoring node, so we can just use:

`kolla_internal_vip_address: "{{ internal_net_name | net_ip }}"`

ansible/overcloud-grafana-configure.yml
```diff
@@ -0,0 +1,40 @@
...
+    - role: stackhpc.grafana-conf
+      grafana_conf_organisation: "{{ grafana_control_plane_organisation }}"
+      grafana_conf_grafana_admin_user: "{{ grafana_local_admin_user_name }}"
+      grafana_conf_grafana_admin_pass: "{{ grafana_admin_password }}"
+      grafana_conf_grafana_dashboard_repo:
+        path: "{{ grafana_monitoring_node_dashboard_repo }}"
+        checkout_path: "{{ grafana_monitoring_node_dashboard_repo_path }}"
+        relative_path: "{{ grafana_monitoring_node_dashboard_path }}"
+      grafana_conf_grafana_url: "http://{{ kolla_internal_vip_address }}:3000"
+      grafana_conf_grafana_datasources: "{{ grafana_datasources }}"
```
markgoddard: Have you checked that this is a noop with the defaults?

kayobe/cli/commands.py
markgoddard: Nice, thanks

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,42 @@
+---
+###############################################################################
+# Grafana configuration.
+
+# Grafana local admin user name. If you are deploying Monasca Grafana this
+# should not conflict with an OpenStack user name.
+grafana_local_admin_user_name: "grafana_admin"
```
markgoddard: Perhaps we should stick with the kolla default here and override in kayobe-config, since monasca will not be in use by all users.

ansible/group_vars/all/grafana
```diff
@@ -0,0 +1,42 @@
...
+
+# Path to git repo containing Grafana dashboards. Eg.
+# https://github.com/stackhpc/grafana-reference-dashboards.git
+grafana_monitoring_node_dashboard_repo:
+
+# Path to which Grafana dashboards will be cloned to a monitoring node
+grafana_monitoring_node_dashboard_repo_path: "{{ source_checkout_path ~ '/grafana' }}"
```
markgoddard: nit: grafana-dashboards?

markgoddard (changes requested):
Looking better.

ansible/group_vars/all/grafana
dougszumski: Done

ansible/group_vars/all/grafana
dougszumski: Done

ansible/group_vars/all/grafana
dougszumski: Done

ansible/group_vars/all/grafana
dougszumski: Done

ansible/overcloud-grafana-configure.yml
dougszumski: Done - thanks for the pointer.

ansible/overcloud-grafana-configure.yml
dougszumski: It runs and doesn't change anything. There are a few tasks in the role which read existing state.

markgoddard (approved):
Looks good.

dougszumski:
Thanks for reviewing!

ansible/overcloud-grafana-configure.yml
```diff
@@ -0,0 +1,41 @@
...
+  roles:
+    - role: stackhpc.grafana-conf
+      grafana_conf_organisation: "{{ grafana_control_plane_organisation }}"
+      grafana_conf_grafana_admin_user: "{{ grafana_local_admin_user_name }}"
+      grafana_conf_grafana_admin_pass: "{{ grafana_admin_password }}"
+      grafana_conf_grafana_dashboard_repo:
+        repo: "{{ grafana_monitoring_node_dashboard_repo }}"
+        version: "{{ grafana_monitoring_node_dashboard_repo_version }}"
+        checkout_path: "{{ grafana_monitoring_node_dashboard_repo_checkout_path }}"
+        relative_path: "{{ grafana_monitoring_node_dashboard_repo_path }}"
```
markgoddard, line 39: What happens here with the defaults? Won't we pass in:

```
     grafana_conf_grafana_dashboard_repo:
        repo:
        version:
        checkout_path: "<default path>"
        relative_path:
```

dougszumski: Thanks - that messes up the logic in the ansible role.

markgoddard: An alternative would be to just pass through the same dict format from kayobe.

dougszumski: That would neater. For the sake of saving time I'm tempted to just set a fact to make sure everything is defined in the grafana role.

PR #119, opened 2018-01-31
Add scripting around the development environment

markgoddard:
The following top level scripts are added under dev/:

* overcloud-deploy.sh
* seed-deploy.sh
* seed-hypervisor-deploy.sh

Some amount of configuration is possible via dev/config.sh.

### DIFF
--- a/dev/environment-setup.sh
+++ b/dev/environment-setup.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+
+set -e
+
+# This script can be used to prepare the environment for use with kayobe. This
+# includes setting environment variables and activating the python virtual
+# environment. This script should be sourced rather than executed in a
+# subprocess. e.g. source dev/environment-setup.sh
+
+PARENT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+source ${PARENT}/functions
+
+
+function main {
+    config_init
+    environment_setup
+}
+
+main
--- a/dev/environment-setup.sh
+++ b/dev/environment-setup.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-set -e
+set -eu
+set -o pipefail
 
 # This script can be used to prepare the environment for use with kayobe. This
 # includes setting environment variables and activating the python virtual
@@ -9,7 +10,7 @@ set -e
 
 PARENT="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
-source ${PARENT}/functions
+source "${PARENT}/functions"
 
 
 function main {
--- a/dev/functions
+++ b/dev/functions
@@ -1,6 +1,7 @@
 #!/bin/bash
 
-set -e
+set -eu
+set -o pipefail
 
 # Library of functions for the kayobe development environment.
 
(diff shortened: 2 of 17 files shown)


doc/source/development/automated.rst
```diff
@@ -0,0 +1,151 @@
...
+-----
+
+SSH into the Vagrant VM::
+
+    vagrant ssh
+
+Run the ``dev/overcloud-deploy.sh`` script to deploy the OpenStack control
+plane::
+
+    /vagrant/kayobe/dev/overcloud-deploy.sh
```
yankcrime: This path should be `/vagrant/dev/overcloud-deploy.sh`

markgoddard: Good spot.

dev/environment-setup.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/bash
+
+set -e
```
dougszumski: Add ```set -u``` ? Same comment for other scripts.

dev/functions
```diff
@@ -0,0 +1,195 @@
...
+    else
+        echo "Pulling overcloud container images"
+        kayobe overcloud container image pull
+    fi
+
+    echo "Deploying containerised overcloud services"
+    kayobe overcloud service deploy
+
+    echo "Performing post-deployment configuration"
+    source ${KOLLA_CONFIG_PATH:-/etc/kolla}/admin-openrc.sh
```
dougszumski: nit: add double quotes in-case someone puts a space in their config path? Same for other vars.

doc/source/development/vagrant.rst
```diff
@@ -0,0 +1,55 @@
...
+* VirtualBox
+* VMWare Fusion
+
+The VM is configured with 4GB RAM. It has a single private network in addition
+to the standard Vagrant NAT network.
+
+Preparation
+===========
+
+First, ensure that Vagrant is installed and correctly configured to use
```
dougszumski, line 21: Presumably we could also use VMWare Fusion as noted above? Should we mention that a Vagrant provider should be installed and configured at this stage as well?

doc/source/development/vagrant.rst
```diff
@@ -0,0 +1,55 @@
...
+virtual box. Also install the following vagrant plugins::
+
+    vagrant plugin install vagrant-vbguest
+    vagrant plugin install vagrant-reload
+
+Note: if using Ubuntu 16.04 LTS, you may be unable to install any plugins. To
```
dougszumski: FWIW, i found another workaround which didn't require using the upstream version:
```sudo sed -i'' "s/Specification.all = nil/Specification.reset/" /usr/lib/ruby/vendor_ruby/vagrant/bundler.rb```

dougszumski (approved):
Looks good to me. I started setting it up, but I hit an issue with having to sign the vbox kernel module, so I've parked it for now.

dev/environment-setup.sh
dougszumski: I suppose we could also add bashate?

markgoddard: Done and done.

markgoddard: Also pipefail.

dev/functions
markgoddard: Now more quotey.

doc/source/development/vagrant.rst
markgoddard: Done.

PR #120, opened 2018-01-31
Virtualenv parent dir creation, ansible_user_id, ironic rescue interfaces

markgoddard:


dougszumski (approved):
This should be nice to have.

Issue #112, opened 2018-01-12
Overcloud host bifrost network bootstrap issues


markgoddard:
When provisioning bare metal overcloud hosts via kayobe overcloud provision, typically the hosts cannot be SSH'd into for at least 5-10 minutes due to DHCP timeouts. Sometimes the hosts cannot be accessed even after this timeout and must be manually configured via the console before proceeding with the kayobe overcloud host configure command.

markgoddard:
There are a few issues at play here.
- Cloud-init is not reading the `network_data.json` file from the ConfigDrive, so applies a 'fallback' configuration of DHCP on one interface. There is no DHCP server so this is where the timeout comes from.
- Cloud-init does not use the `network_data.json` file under `openstack/latest` due to https://bugs.launchpad.net/bifrost/+bug/1742957.
- The CentOS cloud images since 7.2 (1511) have a bogus name server entry in resolv.conf, which cloud-init does not remove. See https://bugs.centos.org/view.php?id=14369.

markgoddard:
Bifrost issue fixed by https://github.com/stackhpc/bifrost/pull/1 and https://review.openstack.org/533196.

Issue #111, opened 2018-01-12
Seed VM network bootstrap issues


markgoddard:
When provisioning a seed VM via `kayobe seed vm provision`, typically the VM cannot be SSH'd into for at least 5-10 minutes due to DHCP timeouts. Sometimes the VM cannot be accessed even after this timeout and must be manually configured via the console before proceeding with the `kayobe seed host configure` command.

markgoddard:
A couple of issues are at play here.
- Cloud-init expects the network metadata file to be called network_data.json, rather than the network_info.json provided by the jriguera.configdrive role.
- Cloud-init requires all links in network_data.json to have either an ether_mac_address or a name field. We provide neither. The name is most accessible and already known, so we'll use this.

markgoddard:
Fixed by https://github.com/jriguera/ansible-role-configdrive/pull/2 and https://github.com/jriguera/ansible-role-configdrive/pull/3.

markgoddard:
Fixes merged to jriguera.configdrive.

PR #122, opened 2018-02-01
Cloud init disable

diranetafen:
Disable Cloud-init to prevent the regeneration of network interface config

### DIFF
--- a/ansible/group_vars/all/overcloud
+++ b/ansible/group_vars/all/overcloud
@@ -16,6 +16,9 @@ overcloud_groups: >
 # should not be added to the inventory.
 overcloud_group_hosts_map: {}
 
+# To prevent some network issues you can choose to disable cloud-init
+disable_cloud_init: False 
+
 ###############################################################################
 # Overcloud host image configuration.
 
--- a/ansible/roles/disable-cloud-init/handlers/main.yml
+++ b/ansible/roles/disable-cloud-init/handlers/main.yml
@@ -1,0 +1,7 @@
+---
+- name: restart cloud-init daemon
+  systemd: 
+    name: cloud-init
+    state: restarted 
+    daemon_reload: yes
+  become: True
--- a/ansible/roles/disable-cloud-init/tasks/main.yml
+++ b/ansible/roles/disable-cloud-init/tasks/main.yml
@@ -1,0 +1,9 @@
+---
+- name: Disable cloud init service
+  file:
+    path: /etc/cloud/cloud-init.disabled 
+    state: touch
+    mode: "u=rw,g=r,o=r" 
+  notify:
+    - restart cloud-init daemon 
+  become: True
--- a/kayobe/tests/unit/cli/test_commands.py
+++ b/kayobe/tests/unit/cli/test_commands.py
@@ -466,6 +466,7 @@ class TestCase(unittest.TestCase):
                     "ansible/network.yml",
                     "ansible/sysctl.yml",
                     "ansible/disable-glean.yml",
+                    "ansible/disable-cloud-init.yml",
                     "ansible/ntp.yml",
                     "ansible/lvm.yml",
                 ],
--- a/ansible/disable-cloud-init.yml
+++ b/ansible/disable-cloud-init.yml
@@ -1,0 +1,12 @@
+---
+# Cloud-init ‘s searches for network configuration in order of 
+# increasing precedence; each item overriding the previous.
+# In some cases cloud-init reconfigure automatically network interface
+# and cause some issues in network configuration
+- name: Disable Cloud-init service 
+  hosts: overcloud
+  tags:
+    - disable-cloud-init
+  roles:
+    - role: disable-cloud-init
+      when: disable_cloud_init | bool
--- a/etc/kayobe/overcloud.yml
+++ b/etc/kayobe/overcloud.yml
@@ -14,7 +14,7 @@
 #overcloud_group_hosts_map:
 
 # To prevent some network issues you can choose to disable cloud-init
-disable_cloud_init: False 
+#disable_cloud_init:
 
 ###############################################################################
 # Overcloud host image configuration.
--- a/etc/kayobe/overcloud.yml
+++ b/etc/kayobe/overcloud.yml
@@ -13,6 +13,9 @@
 # should not be added to the inventory.
 #overcloud_group_hosts_map:
 
+# To prevent some network issues you can choose to disable cloud-init
+disable_cloud_init: False 
+
 ###############################################################################
 # Overcloud host image configuration.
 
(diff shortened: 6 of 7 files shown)


kayobe/tests/unit/cli/test_commands.py
```diff
@@ -466,6 +466,7 @@ def test_overcloud_host_configure(self, mock_kolla_run, mock_run,
                     "ansible/network.yml",
                     "ansible/sysctl.yml",
                     "ansible/disable-glean.yml",
+                    "ansible/disable-cloud-init.yml"
```
markgoddard: Missing the trailing semicolon - this is why the test is failing.

ansible/roles/disable-cloud-init/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  with_items: "{{ cloud_init_services.stdout_lines }}"
+  become: True
+
+
+- name: Ensure Cloud-init artifacts are removed
+  file:
+    path: "{{ item }}"
+    state: absent
+  with_items:
+    - "/usr/lib/systemd/system/cloud-init@.service"
```
markgoddard: I don't see this file on my system.

ansible/roles/disable-cloud-init/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
...
+  register: cloud_init_services
+  changed_when: False
+
+- name: Ensure Cloud-init services are stopped and disabled
+  service:
+    name: "{{ item.split()[0] }}"
+    state: stopped
+    enabled: no
+  with_items: "{{ cloud_init_services.stdout_lines }}"
+  become: True
```
markgoddard: Also notify the systemd daemon here?

ansible/roles/disable-cloud-init/tasks/main.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: Check for existing CLoud-init systemd services
```
markgoddard: nit: CLoud-init

ansible/roles/disable-cloud-init/handlers/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: Reload systemd daemon
+  command: systemctl daemon-reload
+  become: True
```
markgoddard: nit: you can use http://docs.ansible.com/ansible/latest/systemd_module.html.

markgoddard (changes requested):
Thanks for this Dirane.

I recently spent some time ensuring that cloud-init works well both for the seed and the overcloud hosts. This change would prevent cloud-init from working. At the least we need to put this behind a flag, e.g. `disable_cloud_init: False`. See https://github.com/stackhpc/kayobe/issues/111 and https://github.com/stackhpc/kayobe/issues/112.

Another (smaller) concern I have is that by disabling cloud-init after the first boot, we require that first boot to be successful. I had a few problems with getting glean and cloud-init to work together. Wouldn't a better solution be to disable or uninstall cloud-init in the image so that it never runs?

Finally, have you tried other methods of disabling cloud-init's network config, such as those listed here: http://cloudinit.readthedocs.io/en/latest/topics/network-config.html#disabling-network-configuration? The simplest I think would be to write a file to /etc/cloud/cloud.cfg.d/disable-net.cfg, containing this:

```
network:
  config: disabled
```

ansible/roles/disable-cloud-init/handlers/main.yml
diranetafen: Thanks for the link ! i will use systemd: daemon_reload=yes

ansible/roles/disable-cloud-init/tasks/main.yml
diranetafen: :-) changed

ansible/roles/disable-cloud-init/tasks/main.yml
diranetafen: I have those files in my environment

kayobe/tests/unit/cli/test_commands.py
diranetafen: corrected

ansible/roles/disable-cloud-init/handlers/main.yml
```diff
@@ -1,4 +1,3 @@
 ---
 - name: Reload systemd daemon
-  command: systemctl daemon-reload
-  become: True
+  systemd: daemon_reload=yes
```
markgoddard: Still require become.

markgoddard: nit: prefer YAML syntax:

```
systemd:
  daemon_reload: yes
```

markgoddard (changes requested):

diranetafen:
Hi @markgoddard 
i'm still working on my pull request https://github.com/stackhpc/kayobe/pull/122
And I have a difficulty to understand exactly the problem with it! Because my playbook runs when the overcloud hosts are configured, so the first boot will complete successfully (after provision overcloud).

diranetafen:
I will make some changes in my code and test it based on this article : https://fatmin.com/2017/10/19/how-to-disable-cloud-init-in-a-rhel-cloud-image/

ansible/roles/disable-cloud-init/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Disable cloud init service
+  file:
+    path: /etc/cloud/cloud-init.disabled 
+    state: touch
+    mode: "u=rw,g=r,o=r" 
+  when: disable_cloud_init | bool
```
markgoddard: nit: Could you move this `when` to the disable-cloud-init.yml playbook, like this:

```
role: disable-cloud-init
when: disable_cloud_init | bool
```

ansible/group_vars/all/overcloud
```diff
@@ -16,6 +16,9 @@ overcloud_groups: >
 # should not be added to the inventory.
 overcloud_group_hosts_map: {}
 
+# To prevent some network issues you can choose to disable cloud-init
+disable_cloud_init: False
```
markgoddard, line 20: Please update etc/kayobe/overcloud.yml.

markgoddard (changes requested):
This is looking better, thanks.

ansible/group_vars/all/overcloud
diranetafen: Done !

markgoddard (approved):
I made a small change to etc/kayobe/overcloud.yml to comment out the variable and remove its value. Otherwise, looks good - thanks!

diranetafen:
Ok @markgoddard  thanks

PR #125, opened 2018-02-19
Feature/support configuring grafana

dougszumski:


ansible/roles/kolla-openstack/templates/grafana.ini.j2
```diff
@@ -1,14 +1,5 @@
 # {{ ansible_managed }}
 
-[security]
-; If an OpenStack user with the same name as the admin user logs into
-; Grafana it overwrites user data in the Grafana database, breaking
-; the local admin account, and preventing admin API calls to Grafana. To
-; reduce the chance of this happening we rename the local admin user here.
-; Note that this only affects the Monasca fork of Grafana.
-;admin_user = admin
-admin_user = {{ grafana_local_admin_user_name }}
```
markgoddard, line 10: Should we set this in globals.yml.j2 in the kolla-openstack role? Would save needing to set it in kayobe-config. Could make it dependent upon kolla_enable_grafana and grafana_local_admin_user_name being set.

markgoddard (approved):

ansible/overcloud-grafana-configure.yml
```diff
@@ -8,6 +8,30 @@
...
+      os_openstackclient_venv: "{{ venv }}"
+
+  tasks:
+    - name: Look up Monasca control plane project ID
+      shell: >
+        source {{ venv }}/bin/activate &&
+        openstack project show monasca --format json --column id
+      register: monasca_project_show
+      changed_when: False
+      environment: "{{ openstack_auth_env }}"
```
markgoddard, line 30: @ktibi asked if he could use this feature. As it stands, it depends on the monasca project existing. I think ideally we'd make these two tasks dependent upon a `kolla_enable_monasca` variable. It might be a bit misleading to add that as it stands as it wouldn't do anything. Alternatively we could add `failed_when: False` to ignore errors. What do you think?

markgoddard (changes requested):

markgoddard:
Also, ansible tests are failing for some reason.

ansible/overcloud-grafana-configure.yml
dougszumski: Hmm. I like the idea of being explicit with ```kolla_enable_monasca```. If I put a big comment near it explaining that it's work in progress hopefully it shouldn't be too misleading?

markgoddard: Works for me, and is more future proof.

ansible/roles/kolla-openstack/defaults/main.yml
```diff
@@ -237,6 +237,15 @@ kolla_enable_magnum:
...
 kolla_extra_magnum:
 
+###############################################################################
+# Monasca configuration.
+
+# NOTE: Support for Monasca in Kolla is a work in progress. This currently
+# relies on an out-of-band Monasca installation.
+
+# Whether to enable Monasca.
+kolla_enable_monasca:
```
markgoddard, line 247: To pass this through to kolla-ansible, it needs to be added to the list in vars/main.yml in this role.

ansible/overcloud-grafana-configure.yml
```diff
@@ -18,6 +18,7 @@
   roles:
     - role: stackhpc.os-openstackclient
       os_openstackclient_venv: "{{ venv }}"
+      when: kolla_enable_monasca
```
markgoddard: I think this (and the others below) needs a | bool, as the default "no" evaluates to True.

markgoddard (changes requested):

ansible/overcloud-grafana-configure.yml
dougszumski: Thanks, fixed.

ansible/roles/kolla-openstack/vars/main.yml
```diff
@@ -75,6 +75,11 @@ kolla_openstack_custom_config:
     dest: "{{ kolla_node_custom_config_path }}/manila"
     patterns: "*"
     enabled: "{{ kolla_enable_manila }}"
+  # Monasca.
+  - src: "{{ kolla_extra_config_path }}/monasca"
+    dest: "{{ kolla_node_custom_config_path }}/monasca"
+    patterns: "*"
+    enabled: "{{ kolla_enable_monasca }}"
```
markgoddard: Sorry, I meant the list in kolla-ansible/vars/main.yml. This change will copy config files, but we don't have anything to consume them in kolla-ansible yet.

markgoddard (changes requested):

ansible/roles/kolla-openstack/vars/main.yml
dougszumski: Thanks - that makes more sense. Hopefully it won't be long until we do have something.

dougszumski:
Ready for review.

markgoddard (approved):

PR #129, opened 2018-02-22
Respect Kolla upper constraints in virtual envs

dougszumski:
This resolves the issues I was having at least - it downgraded docker-py in the kolla venv for me.

### DIFF
--- a/ansible/group_vars/all/kolla
+++ b/ansible/group_vars/all/kolla
@@ -271,6 +271,9 @@ kolla_external_fqdn_cert:
 # Whether debug logging is enabled.
 kolla_openstack_logging_debug: "False"
 
+# Upper constraints file for the stable/pike branch of Kolla
+kolla_upper_constraints_file: "https://raw.githubusercontent.com/openstack/requirements/stable/pike/upper-constraints.txt"
+
 ###############################################################################
 # Kolla feature flag configuration.
 
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -18,6 +18,7 @@
   pip:
     name: docker
     state: latest
+    extra_args: "-c {{ kolla_upper_constraints_file }}"
     virtualenv: "{{ virtualenv is defined | ternary(virtualenv, omit) }}"
   become: "{{ virtualenv is not defined }}"
 
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -18,7 +18,7 @@
   pip:
     name: docker
     state: latest
-    extra_args: "-c {{ docker_upper_constraints_file }}"
+    extra_args: "{% if docker_upper_constraints_file %}-c {{ docker_upper_constraints_file }}{% endif %}"
     virtualenv: "{{ virtualenv is defined | ternary(virtualenv, omit) }}"
   become: "{{ virtualenv is not defined }}"
 
--- a/ansible/roles/docker/tasks/main.yml
+++ b/ansible/roles/docker/tasks/main.yml
@@ -18,7 +18,7 @@
   pip:
     name: docker
     state: latest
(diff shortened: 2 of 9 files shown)


dougszumski:
Closes #129

ansible/roles/kolla/tasks/install.yml
```diff
@@ -51,17 +51,22 @@
...
     version: "{{ item.version | default(omit) }}"
     state: "{{ item.state | default('present') }}"
     virtualenv: "{{ kolla_venv }}"
+    extra_args: "-c {{ kolla_upper_constraints_file }}"
   with_items:
-    # In version 2.0.0, docker renamed the docker-py python package to docker.
-    # Kolla requires the docker package rather than the docker-py package.
-    - name: docker-py
-      state: absent
     - name: docker
```
markgoddard: Related: kolla now includes this as a dependency: https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=stable/pike. Also, PyYAML is not required since kolla-ansible was split from kolla.

ansible/group_vars/all/kolla
```diff
@@ -271,6 +271,9 @@ kolla_external_fqdn_cert:
 # Whether debug logging is enabled.
 kolla_openstack_logging_debug: "False"
 
+# Upper constraints file for the stable/pike branch of Kolla
+kolla_upper_constraints_file: "https://raw.githubusercontent.com/openstack/requirements/stable/pike/upper-constraints.txt"
```
markgoddard, line 275: nit: I think it's preferred to use the OpenStack git repos rather than the github mirror:
`https://git.openstack.org/cgit/openstack/requirements/plain/upper-constraints.txt?h=stable/pike`

ansible/roles/docker/tasks/main.yml
```diff
@@ -18,6 +18,7 @@
   pip:
     name: docker
     state: latest
+    extra_args: "-c {{ kolla_upper_constraints_file }}"
```
markgoddard: nit: This should be in the role defaults, and ideally use a docker_ prefix rather than kolla_. You could translate the variable in ansible/docker.yml.

ansible/roles/kolla-ansible/tasks/install.yml
```diff
@@ -58,6 +58,7 @@
   pip:
     requirements: "{{ kolla_ansible_venv }}/requirements.txt"
     state: present
+    extra_args: "-c {{ kolla_upper_constraints_file }}"
```
markgoddard: I think this is failing because it's not in the role defaults.

ansible/roles/kolla/tasks/install.yml
```diff
@@ -51,17 +51,22 @@
...
+    state: absent
+    virtualenv: "{{ kolla_venv }}"
+
 - name: Ensure required Python packages are installed
   pip:
     name: "{{ item.name }}"
     version: "{{ item.version | default(omit) }}"
     state: "{{ item.state | default('present') }}"
     virtualenv: "{{ kolla_venv }}"
+    extra_args: "-c {{ kolla_upper_constraints_file }}"
```
markgoddard: Should be in the role defaults.

markgoddard (changes requested):

ansible/roles/kolla-ansible/tasks/install.yml
markgoddard: I think this will still fail because the variable will be none in the test. We either need to provide a sensible value in the default, or set a value in the test.

markgoddard (changes requested):

dougszumski:
Thanks - fixing it. Will remove WIP flag when ready for re-review.

dougszumski:
Ready for re-review. As an alternative to just omitting the upper constraints when they're not specified I suppose I could add the master branch constraints as a default.

markgoddard (approved):
LGTM, thanks.

Issue #130, opened 2018-02-23
Nested list of storage groups causes templating of overcloud inventory to fail


dougszumski:
# Overview

The storage group mapping from kolla-ansible to kayobe added in PR #84 uses a nested list. This causes the generation of the overcloud inventory to fail.

# Steps to reproduce

For example, this was seen whilst running:

```kayobe overcloud service reconfigure -kt haproxy```

# Expected results

Overcloud inventory file is generated

# Actual results

```
TASK [kolla-ansible : Ensure the Kolla overcloud inventory file exists] **********************************************************************************************************************************************************************
[...]

PR #131, opened 2018-02-23
Issue #130: De-nest storage groups

dougszumski:
Addresses issue #130


markgoddard (approved):
Thanks Doug, I also hit this and it was a pain to debug. Sorry for not pushing the patch quickly enough.

Issue #138, opened 2018-03-05
Tags don't work with kayobe overcloud post configure


markgoddard:
# Steps to reproduce

kayobe overcloud post configure --tags provision-net

# Expected results

The provision-net playbook tasks are executed.

# Actual results

The provision-net playbook tasks are skipped.

PR #139, opened 2018-03-05
Fix use of tags with overcloud post configure

markgoddard:
When specifying tags with 'kayobe overcloud post configure --tags <tags>', the
expected plays typically do not run. This seems to be due to an ansible bug,
where if a dynamically created group is referenced before creation, it will
remain empty. See https://github.com/ansible/ansible/issues/20360.

This change works around this issue by modifying the name of the dynamic groups
used in the post configure playbooks to be unique to each playbook.

Fixes: #138


dougszumski (approved):
LGTM, ```monitoring_with_grafana_enabled_ appears in only one playbook```.

PR #140, opened 2018-04-26
Enable iPXE boot in ironic

jovial:
a new variable, kolla_enable_ironic_ipxe is added to control whether or not
iPXE is used as the network boot program in ironic. You can set this is in
etc/kayobe/kolla.yml, e.g:

	kolla_enable_ironic_ipxe: True

This depends on:

- stackhpc/kolla#6
- stackhpc/kolla-ansible#29


markgoddard (changes requested):
Changes look reasonable, but we now use gerrit for kayobe development. You should make sure that your kayobe repo is pointing upstream also.

jovial:
pull request moved to: https://review.openstack.org/#/c/565469/

PR #144, opened 2018-06-22
Enable iPXE boot in ironic

jovial:
a new variable, kolla_enable_ironic_ipxe is added to control whether or not
iPXE is used as the network boot program in ironic. You can set this is in
etc/kayobe/kolla.yml, e.g:

  kolla_enable_ironic_ipxe: True

Change-Id: I7da15e7ceea580e5b7bee7fc732351cbec666f4c
Story: 2001949
Task: 15068
Depends-On: I08fca1864a00afb768494406c49e968920c83ae7


markgoddard (approved):

markgoddard:
This is stackhpc/queens, guess this needs to go into stackhpc/pike too?

PR #146, opened 2018-06-26
Adds a separate admin network

jovial:
The admin network is intended for remote admin access to
the overcloud hosts e.g SSH. If admin_oc_net_name is not set
it will default to provision_oc_net_name for backwards
compatability.

Story: 2002096
Task: 19774

Change-Id: Ib04bbc07f97218d7503000cc363624c60c287822


markgoddard (commented):
This one seems to have the workaround patch in too.

jovial:
I've removed the cloud-init workaround now

markgoddard (approved):

PR #158, opened 2018-08-31
Fix kayobe overcloud service config save

jovial:
If Kolla Ansible config files are not readable by the kayobe (stack)
user, the command will fail. This is fixed by using become.

Also improves the command output, by showing only the path of each file
rather than the full stat result.

Change-Id: I466e6a75822a1b2d2f91c9fadc9730c2cdb5bba0
TrivialFix


markgoddard (approved):

markgoddard:
Can you also cherry pick to stackhpc/queens?

PR #163, opened 2018-09-11
Feature/monasca pike

dougszumski:


ansible/roles/kolla-openstack/vars/main.yml
```diff
@@ -95,6 +100,21 @@ kolla_openstack_custom_config:
...
+    ignore:
+      # These are templated by kayobe, so don't remove them.
+      - log-transformer.conf
+      - log-persister.conf
+      - log-metrics.conf
+      - notification.conf
+      - api-config.conf
+      - log-api.conf
+      - thresh-config.yml
+      - persister.conf
```
markgoddard: Is there a reason to template these specifically? If you remove this ignore attribute then these files will be templated, and you don't need to do the extra stuff, or define a template in kayobe.

Normally we do this if kayobe wants to inject some default config of its own.

ansible/group_vars/all/kolla
```diff
@@ -168,6 +168,8 @@ overcloud_container_image_regex_map:
     enabled: True
   - regex: memcached
     enabled: True
+  - regex: monasca
+    enabled: "{{ kolla_enable_monasca | bool }}"
```
markgoddard: Should we use this instead: https://review.openstack.org/598520? This will break any binary environments with monasca enabled.

markgoddard (changes requested):

markgoddard:
Any reason not to do this upstream first and backport?

ansible/group_vars/all/kolla
dougszumski: The StackHPC ones have this set so it shouldn't break them: https://github.com/SKA-ScienceDataProcessor/alaska-kayobe-config/blob/alaska-prod/etc/kayobe/kolla/globals.yml#L7

ansible/roles/kolla-openstack/vars/main.yml
dougszumski: No particular reason yet, I'll update it to use the simpler approach. Thanks.

ansible/group_vars/all/kolla
dougszumski: Now I see that patch is already backported :) A customer wants them urgently in pike hence no upstream patch first.

dougszumski:
Cherry-picked from upstream. Need to remember to use ```cherry-pick -x```...

markgoddard (approved):

PR #166, opened 2018-09-17
Ansible < 2.4 compatability

jovial:
include_tasks is not available in Ansible < 2.4. It seems the
behaviour is slightly different when using include - hence the other
fixups

### DIFF
--- a/ansible/roles/ipa-images/tasks/set-driver-info.yml
+++ b/ansible/roles/ipa-images/tasks/set-driver-info.yml
@@ -32,7 +32,7 @@
   when:
     - item.Name is not none
     - item.Name not in groups["baremetal-compute"]
-  with_items: "{{ ipa_images_ironic_node_list.stdout | from_json }}"
+  with_items: "{{ ipa_images_ironic_node_list.stdout | default('{}') | from_json }}"
 
 - name: Set fact containing filtered list of nodes
   set_fact:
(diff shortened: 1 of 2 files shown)


ansible/roles/ipa-images/tasks/set-driver-info.yml
```diff
@@ -32,7 +32,7 @@
   when:
     - item.Name is not none
     - item.Name not in groups["baremetal-compute"]
-  with_items: "{{ ipa_images_ironic_node_list.stdout | from_json }}"
+  with_items: "{{ ipa_images_ironic_node_list.stdout | default('{}') | from_json }}"
```
markgoddard, line 35: Is this Ansible 2.4 specific?

markgoddard (approved):

ansible/roles/ipa-images/tasks/set-driver-info.yml
jovial: I guess we this upstream aswell. Although I'm surprised that it gets evaluated given the `when` condition on the include. Is there a difference in the behaviour of `when` between `include` and `include_tasks`?

markgoddard: I'm not sure about include, but include_tasks will completely ignore the tasks in the file if the condition is false. Include might be more like import_tasks, which is a static include.

PR #180, opened 2018-10-03
Remove extra configuration for InfluxDB

dougszumski:
InfluxDB config file merging isn't supported in Kolla-Ansible because
it uses 'nested sections' which aren't supported by merge_configs. If
no override file is specified, Kayobe will write out an empty config
file which will then be used as the InfluxDB config file, breaking
InfluxDB. To prevent that happening this change removes the extra
config in Kayobe. It also fixes the directory to which the 'glob'
collected config is copied to, as Kolla-Ansible doesn't look for
Influxdb config files in the influxdb folder.

Story: 2003951
Task: 26868

### DIFF
--- a/ansible/kolla-openstack.yml
+++ b/ansible/kolla-openstack.yml
@@ -106,7 +106,6 @@
             - { name: glance, file: glance.conf }
             - { name: grafana, file: grafana.ini }
             - { name: heat, file: heat.conf }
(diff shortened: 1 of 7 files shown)


markgoddard (approved):
Happy to take this downstream if it gets you moving faster, but suggested what (I think) is a more consistent fix upstream.

PR #185, opened 2018-10-05
Fix inspector dnsmasq interface name

markgoddard:
Kolla-ansible uses the variable 'ironic_dnsmasq_interface' to denote the
interface for the ironic inspector dnsmasq service. Previously kayobe was
incorrectly using the variable 'inspector_dnsmasq_interface', which caused
kolla-ansible to ignore it and use the API interface instead. This patch fixes
that.

Change-Id: I733a84759cd03b62659dbf2d7027b7be9e42e818


dougszumski (approved):
Looks good - couldn't see any more instances with grep.

PR #216, opened 2021-10-25
Wallaby: Build overcloud host image directly with DIB

cityofships:
Backporting https://review.opendev.org/c/openstack/kayobe/+/772609


ansible/group_vars/all/dib
```diff
@@ -0,0 +1,54 @@
+---
```
markgoddard: There was a question in the upstream review about naming. It could be quite confusing to change filenames and variable names in our fork. I'd suggest that we at least agree on an approach, and submit it to upstream review, before merging here.

markgoddard (changes requested):

ansible/group_vars/all/dib
cityofships: Now merged upstream

markgoddard (approved):

PR #232, opened 2022-03-24
Support running libvirt as a host daemon

markgoddard:
1. libvirt: deploy libvirt on the host

In some cases it may be desirable to run libvirt daemon on the host. For example, when mixing host and container OS distributions.
    
This change makes it possible to disable the nova_libvirt container, by setting `kolla_enable_nova_libvirt_container` to false.
    
The stackhpc.libvirt-host role is used in order to install and configure a libvirt daemon on compute hosts when `kolla_enable_nova_libvirt_container` is false.

Change-Id: I73fef63fb886a9d543d2f4231fb009523495edb3
(cherry picked from commit c4b74f48014969759d3783ee895f0a7b9cc68a1d)

2. libvirt: support SASL authentication
    
Adds support for SASL authentication of libvirt TCP and TLS connections when using a compute host libvirt daemon.
    
In line with the dependent Kolla Ansible patch, we enable SASL by default, and use DIGEST-MD5 with TCP and SCRAM-SHA-256 with TLS.

Change-Id: Id3972c24022aeb6421494c3cccdc8e7cbce802e6
(cherry picked from commit f4493e41ff6383dba8967b881362bee2cb7a5794)


jovial (approved):
Worked for me

PR #237, opened 2022-05-05
Add more control to defining kolla_ansible_user

Aethylred:
This adds variables that allows setting home directory, UID, and GID for the Kolla ansible user used by Kayobe


markgoddard:
Thanks @Aethylred. This change should first go upstream via gerrit to the master branch. Is that something you can pick up?

I need to close this PR, since this repository is being recreated for $reasons.

